PR #9, opened 2013-11-22
Disable auto-loading of Sidekiq web-ui-extension

felixbuenemann:
This allows to use sidekiq-cron without the web-ui-integration.

Now you need to explicitly load the cron extension in your routes:
```ruby
require 'sidekiq/web'
require 'sidekiq/cron/web'
mount Sidekiq::Web => '/sidekiq'
```


ondrejbartas:
resolved in #9

PR #36, opened 2015-05-07
In timezone

ondrejbartas:


ondrejbartas:
This is not good idea. Rails zone can be changed so it can be very tricky to set correct time for all users. 

UTC is only one time, no problem with this. You need to calculate first, when job needs to run.

PR #37, opened 2015-05-17
Fix a test that depends on current time

sue445:
When run test from 0:00 to 2:00 every day, always fail a test.

# Example
When run test at 2015/5/18 0:30, raised this error.

```
     FAIL (0:00:00.165) test_: Cron Job cron test should return previous day.
          <"2015-05-17-02-01-00"> expected but was
          <"2015-05-18-02-01-00">.
```

 I think expected date is wrong


sue445:
@ondrejbartas Cause of test failed is that sidekiq is not support MRI 1.9. 
Please merge #38 earlier :bow:

sue445:
:heart:

PR #38, opened 2015-05-17
Don't test on ruby 1.9.3

sue445:
Because Sidekiq does not already support MRI 1.9

https://github.com/mperham/sidekiq/blob/v3.3.4/lib/sidekiq.rb#L3


sue445:
:heart:

PR #49, opened 2015-07-22
Enable disabled question

tfluehmann:
Added the methods to ask the job if it is enabled? or disabled?


lib/sidekiq/cron/job.rb
```diff
@@ -265,6 +265,16 @@ def enable!
         @status = "enabled"
         save
       end
+      
+      def enabled?
+        return @status == "enabled"
```
ondrejbartas: why explicit return?

test/unit/job_test.rb
```diff
@@ -149,7 +157,6 @@
       job = Sidekiq::Cron::Job.new('klass' => 'UnknownCronClass')
       assert_equal job.message, {"class"=>"UnknownCronClass", "args"=>[], "queue"=>"default"}
     end
-
```
ondrejbartas: please return this space back.

ondrejbartas:
If you fix y comments I will merge it.

tfluehmann:
Done, I am using always the explicit return way, because I like to see always what a method returns.
But if it is best practice in this project I don't mind.
Thx for merging.

test/unit/job_test.rb
```diff
@@ -59,6 +59,14 @@
     it "have destroy method" do
       assert @job.respond_to?(:destroy)
     end
+    
+    it "have save method" do
+      assert @job.respond_to?(:enabled?)
+    end
+    
+    it "have save method" do
```
ondrejbartas: wrong naming of test

test/unit/job_test.rb
```diff
@@ -59,6 +59,14 @@
     it "have destroy method" do
       assert @job.respond_to?(:destroy)
     end
+    
+    it "have save method" do
```
ondrejbartas: wrong naming of test

lib/sidekiq/cron/job.rb
```diff
@@ -265,6 +265,16 @@ def enable!
...
+      
+      def enabled?
+        @status == "enabled"
+      end
+      
+      def disabled?
+        !enabled?
+      end
+      
+
```
ondrejbartas: too many whitespaces

tfluehmann:
How is it now?

PR #50, opened 2015-07-25
add csrf tag

antulik:
This pull request fixes compatibility with sidekiq '>= 3.4.2'

Sidekiq recently changed how it works so it broke sidekiq-cron https://github.com/mperham/sidekiq/pull/2422

### DIFF
--- a/lib/sidekiq/cron/views/cron.erb
+++ b/lib/sidekiq/cron/views/cron.erb
@@ -35,19 +35,24 @@
           <td style="<%= style %>">
             <% if job.status == 'enabled' %>
               <form action="<%= root_path %>cron/<%= CGI.escape(job.name).gsub('+', '%20') %>/enque" method="post">
+                <%= csrf_tag %>
                 <input class='btn btn-xs pull-left' type="submit" name="enque" value="<%= t('EnqueueNow') %>"/>
               </form>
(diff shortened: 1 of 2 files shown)


ondrejbartas:
Do you have any idea what will happen when you use Sidekiq web without rails? BTW do you know that we are using Sinatra not Rails? 
Last question why do you need CSRF protection on web which should be accessible only for admins/developers?

antulik:
@ondrejbartas No, I have not tested without rails myself. 

But original Sidekiq web already has that change everywhere https://github.com/mperham/sidekiq/blob/master/web/views/retry.erb#L30 so it should work.

And as per comment https://github.com/mperham/sidekiq/pull/2422#issuecomment-118977854
```
This will break any 3rd party Web UI extensions which try to process POSTs. They will need to update their forms to include the same hidden token input.
```

antulik:
`csrf_tag` is a sinatra helper (not rails)
https://github.com/mperham/sidekiq/commit/5adbc24cbeef10b05f5737e96bb71e2b0b6d0595

siruguri:
CSRF protection is needed even if the route is "protected" by a session cookie based check, which is a common method that most developers will utilize - those session cookies will be part of the trust established with the client that the CSRF exploit is targeting.

ondrejbartas:
resolved in #53 
I needed to add guard so if there is no `csrf_tag` defined it will not break.

This fix is for older sidekiq versions.

PR #56, opened 2015-07-31
Add an active_job property to cron job config.

Junyulive:
In our project, we host the sidekiq-web in a separated project with the actual worker project. Add the active_job property to config will allow the sidekiq-web dashboard to enqueue the cron job with active_job wrapper.


ondrejbartas:
Hi, can you write some test for this?

Junyulive:
Allow set active_job property with string value ("true", "yes") and number (1). 
Add all unit test cases.

PR #73, opened 2015-11-02
Add fix for jobs to not enqueue if they're already past

camkidman:
### The Bug

We discovered that our crons were running at strange times after our deploys, and isolated it down to this: sidekiq-cron will check if jobs have been previously run, but if that data is saved in sidekiq, there will be no history of a job once the server spins up for the first time. This can be a problem for rapid deployment.

### Our Use Case

We have a cron that runs once a month (or is supposed to), but we found out that it was running pretty close to each time we deployed on that specific server and believe it's isolated to the reason above.

### What This PR Does

This PR verifies that when a job is caught in the poller, it hasn't already passed it's scheduled run window. The poller will run every 10 seconds and if a job comes in that is more than a minute late, it won't `enque!`


camkidman:
Investigating the failing tests.

camkidman:
Looks like the tests actually fail in master as well, and the tests that fail locally are different than what fails in Travis.

camkidman:
Still looking into the tests, but it turns out it the root of the problem wasn't the history-erasing issue like I thought, it has something to do with a race condition. When I queue up a job 10_000 times it runs even when it's not supposed to.

mmmries:
Any interest in merging this kind of failsafe? We find it very useful in our production system where we have a monthly job that we don't want to kick off each time redis gets cleared.

camkidman:
It looks like the most recent sync with the changes in master fixed the failing specs. Also, I can confirm this did fix our issue for certain.

mmmries:
@ondrejbartas we have been using this PR in one of our production systems for over a month now and it has been working well for us. Is there any other evidence we can provide for why this change is valuable?

ryanbjones:
bump, would love to see this fixed

ondrejbartas:
bumped to new version `0.4.2`

ondrejbartas:
thank you guys...

PR #77, opened 2015-11-17
Sidekiq 4 support

johnathanludwig:
I've refactored the poller to work with the recent Sidekiq 4 changes. It is now inheriting from the Sidekiq poller which makes it more straightforward. Its possible that sidekiq changes could break it in the future, but I thought this would be better than duplicating the logic, especially since the gem is already pretty heavily dependent on the Sidekiq classes.

This drops support for prior versions of Sidekiq. I didn't have any ideas for a clean way to support them with the difference in dependencies. This update could be released as a major version and only support Sidekiq 4 going forward. Otherwise let me know how you'd like to support them both.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,7 +1,8 @@
 source 'https://rubygems.org'
 
-gem 'sidekiq', '>= 4.0.0.pre2'
+gem 'sidekiq', '>= 4.0.0'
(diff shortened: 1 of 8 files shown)


ondrejbartas:
Thank you for your PR, it looks good.

Only way is to release major version and specify sidekiq dependency. I will update readme to notify users about these changes

PR #83, opened 2015-12-14
job.enque! should not raise exception when class is unknown

lepfhty:
We are running sidekiq-cron in an environment where it should enqueue jobs whose class definition exists outside of its own codebase. Sidekiq workers running on a remote host will read this queue and perform the jobs.

The ActiveJob check tries to constantize the class name, then raises an exception if the class doesn't exist. (If you agree) I think the job definition should still be enqueued, and class lookups should be resolved by the workers instead of by sidekiq-cron

### DIFF
--- a/lib/sidekiq/cron/job.rb
+++ b/lib/sidekiq/cron/job.rb
@@ -44,7 +44,13 @@ module Sidekiq
       def enque! time = Time.now
         @last_enqueue_time = time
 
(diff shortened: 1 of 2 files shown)


ondrejbartas:
@lepfhty thank you

ondrejbartas:
bumped to new version `0.4.2`

PR #86, opened 2016-01-05
Use pipeline to fetch all jobs at once

stormsilver:
In my testing this had a massive perfomance benefit for high-latency environments
with lots of cron jobs.

For my production environment with 1350 cron jobs, it took 287s to run `Job.all` - almost five minutes. By removing the unnecessary existence check and pipelining
the `hgetall` commands, this sped up to under a second.


brodock:
@stormsilver wow, do you see any potential drawback?

stormsilver:
Nope. We have been running this in production for 6 weeks with no problems.

PR #109, opened 2016-08-09
Only update the enqueued key on save if there isn't already one

MMartyn:
I noticed that if a cron job is resaved within ~30 seconds of when it supposed to be enqueued next it won't be enqueued. The resave in my case is a result of the application being restarted and the initializers being triggered. The failure to enqueue appears to be caused by line 417 in job#save. 

I am curious about the need to set the job_enqueued_key on save, but my assumption is so that new jobs aren't immediately queued. Let me know if my assumption is incorrect.

Thanks.

### DIFF
--- a/lib/sidekiq/cron/job.rb
+++ b/lib/sidekiq/cron/job.rb
@@ -414,7 +414,7 @@ module Sidekiq
 
           #add information about last time! - don't enque right after scheduler poller starts!
           time = Time.now
-          conn.zadd(job_enqueued_key, time.to_f.to_s, formated_last_time(time).to_s)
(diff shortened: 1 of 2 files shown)


ondrejbartas:
@MMartyn your assumption is correct. And your code looks good, but travis is failing. I would like to see this code merged, can you update code so it is not failing?

Thanks.

MMartyn:
@ondrejbartas Sure. I am not sure why it is failing in travis as it passes locally, but I'll see if I can track it down.

MMartyn:
@ondrejbartas Not sure if you get notified. Just in case you don't travis passed. Thanks.

ondrejbartas:
@MMartyn 👍

PR #110, opened 2016-08-15
[FIX] Fix a tpyo on the cron dashboard for enabling a job

timminkov:
screen shot 2016-08-15 at 12 18 27 am

Randomly noticed the above. Button should read "Enable", since it re-enables a job.


ondrejbartas:
👍 thanks

PR #121, opened 2016-11-30
Fix rufus-scheduler 3.3.0 ZoTime operation bug.

Junyulive:
ERROR: CRON JOB: can't convert Rufus::Scheduler::ZoTime into an exact number.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,7 +1,7 @@
 source 'https://rubygems.org'
 
 gem 'sidekiq', '>= 4.2.1'
-gem 'rufus-scheduler', '>= 2.0.24'
+gem 'rufus-scheduler', '>= 3.3.0'
 gem 'redis-namespace', '>= 1.5.2'
 
 group :development do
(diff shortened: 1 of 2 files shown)


tisba:
Thanks for the quick fix! We were running into this issue as well.

Maybe only slightly related: We are using Rollbar to collect and aggregate exceptions (for web requests and sidekiq jobs etc). However the sidekiq-cron errors, like this one, was only logged. Is there a way to hook into the error handling?

ondrejbartas:
@tisba sidekiq-cron is using original logger from Sidekiq.logger so if you are not getting any logs it means that you are not getting logs from Sidekiq as well

tisba:
@ondrejbartas hmm okay, I'll take another look. maybe I'm mixing things up then. thanks!

ondrejbartas:
@tisba hope you will find a way, or if you find that sidekiq-cron is doing something wierd pls let us know.

tisba:
Sure, will do!

tisba:
Sorry for the noise, I think there might be indeed room for improvement :) I've created a new issue, because I think this is a bit out of scope for this issue here, see https://github.com/ondrejbartas/sidekiq-cron/issues/123.

PR #134, opened 2017-01-28
Fix grammar in readme

chuchuva:


ondrejbartas:
@chuchuva thanks for your PR. I am from Prague so enaglish is my second languague :(

Issue #142, opened 2017-03-21
Enqueue via web backend is enqueuing jobs with wrong timezone


lichtamberg:
Everytime I enqueue a job in the backend, I can see that the next schedule is in about an hour (59 minutes) - which is wrong (should be queued immediately, not in one hour duration).

As far as I know, you specify the enqueuing time by calling Time.now in your job.rb (which gives back a time in the current time zone) - but somehow this is not matching as it seems (our server is running +0100, rails itself is configured to use UTC).

Shouldn't this be Time.zone.now in your job.rb?
Do you have n hint for me? 
Please correct me if I'm wrong!

lichtamberg:
I've attached a screenshot here.. You can see that the job should run each minute, but is enqueued 1 hour too late :)

![bildschirmfoto 2017-03-22 um 10 52 43](https://cloud.githubusercontent.com/assets/10075/24191972/bfd4a3f6-0eed-11e7-8d1f-a3418b77647d.png)

ondrejbartas:
released new version 0.5.0 and btw thanks for fix

lichtamberg:
You are welcome :)

PR #144, opened 2017-03-22
UTC timezone fix

lichtamberg:
See https://github.com/ondrejbartas/sidekiq-cron/issues/142

This changes fix my mentioned problems, but 2 tests are broken... Maybe someone can take a look at it? 
My knowledge in this gem is not deep enough at the moment.


ondrejbartas:
@lichtamberg thanks for your fix

Issue #147, opened 2017-04-10
Scheduled jobs sometimes don't run


jpserra:
Hi @ondrejbartas 

I have been struggling with this issue for a while now. When we have a bunch of scheduled jobs (various different `crons`) It happens that randomly some jobs don't run. 

When looking into the `Job` class, I found [this](https://github.com/ondrejbartas/sidekiq-cron/blob/master/lib/sidekiq/cron/job.rb#L546) line that I seem to be struggling to understand.

My question is, shouldn't the hardcoded `60` seconds used in `not_past_scheduled_time?` method be the cron period (in seconds) of that particular job instead?

I have seen a lot of forks that the only thing they do is remove the `not_past_scheduled_time?` validation so I believe there's several people with the same issue so It'd be great if you could clarify this.

Thank you!

ondrejbartas:
@jpserra Hi, which version of sidekiq do you use? It can depends on some logic of poller and intervals

jpserra:
I'm using sidekiq-cron (0.4.5) and sidekiq  4.2.10 and i still get ser some jobs not getting enqueued when they're supposed to because the poller takes foo long to check all jobs (more than 60s) causes the problem

ondrejbartas:
@jpserra how many jobs you have?

ondrejbartas:
the whole process of checking enquation of jobs should take jus miliseconds

ondrejbartas:
try to set: `Sidekiq.options[:poll_interval_average]` to 30s

ondrejbartas:
Sidekiq in newer versions changed way how they schedule jobs based on number of workers a some magic

jpserra:
Hi @ondrejbartas 

I start having this issue when I have around 4k jobs scheduled for the same cron.

This is not something necessarily related to sidekiq latests versions, I was having the same issue with `sidekiq-cron (0.4.2)` and `sidekiq (4.1.4)` but since that polling logic was already present in 4.1.4 I'll give that a try, thanks!

jpserra:
Changing Sidekiq's `poll_interval_average` didn't solve the problem either. The only way i manage to have all my schedules enqueued is either adding more dynos or simply removing the `not_past_scheduled_time?` since the issue is that it takes more than `60` seconds to enqueue all the 4k jobs. 

Do you foresee any unexpected behaviour if that verification is removed?

ondrejbartas:
[...]

PR #150, opened 2017-04-13
Fix poller to not adding all jobs and given period of time

ondrejbartas:
When having lot of jobs to schedule, time can go and 60s window is too small to enqueue them all.
Remember time of poller enqueue and then use it for all jobs


ondrejbartas:
#147

PR #162, opened 2017-05-25
Update README.md

arthurbryant:
Add cron scheduler explanation. I am not familiar with cron setting and it took me some time to figure out what the (* * * * *) means! So I added it after I figured it out.


ondrejbartas:
@arthurbryant thanks for update 👍

PR #170, opened 2017-08-18
Prevent disclosure of sensitive data

dimiii:


ondrejbartas:
@dimiii btw this will remove these sensitive data completely from messages that are passed to schedule, wouldn't be this better to do only for frontend?

lib/sidekiq/cron/job.rb
```diff
@@ -292,7 +292,7 @@ def initialize input_args = {}
         elsif @klass
           message_data = {
             "class" => @klass.to_s,
-            "args"  => @args,
+            "args"  => hide(@args, args["secret_args"] || [])
```
ondrejbartas: what about using this hide only for frontend?

dimiii: Sorry, I did not dig to much into the code. So, messages are used not only for presentation, but also for processing, right? I should ponder on this PR more.

dimiii:
@ondrejbartas Updated readme and fixed corruption of message. Sorry for neglectful first approach.

dimiii:
@ondrejbartas Ok, real run shows that there is something I should do, probably related to persistence of `secret_args`, i.e. despite passing of tests, PR is not ready yet.

ondrejbartas:
@dimiii not it looks better. For saving data to redis (persistance) we use `to_hash` method on job
https://github.com/ondrejbartas/sidekiq-cron/blob/master/lib/sidekiq/cron/job.rb#L371-L386

so when you add your `secret args` etc then it will be populated to and back from redis

dimiii:
@ondrejbartas That was my first option. But eventually, I decided on storing of masked_message (commit 90e4c9f with rewriting of history). Sure, if you think that persistence of `secret_args` is a better option, I can redo it.

ondrejbartas:
@dimiii I dont like to store just duplicates of data, when you actually don't need ti.  

Please store secret_args and let masked_message be readable on the fly just from fronted 👍

dimiii:
Done. 

I found that work with `@message` is a bit tricky - sometimes it's a String and sometimes a Hash. That led to some mess when we cast to string https://github.com/ondrejbartas/sidekiq-cron/pull/170/commits/5cb12fbbae761887918b0dc8bf31decc1a5986a7#diff-483d185b400e6d37068b87588d35297fL39 but forgot to do same for https://github.com/ondrejbartas/sidekiq-cron/pull/170/commits/5cb12fbbae761887918b0dc8bf31decc1a5986a7#diff-483d185b400e6d37068b87588d35297fL41


Anyway, as now on front-end the `masked_message` is used, I made `message` only a Hash. Also, I discovered that `@args` is never a `String`, so I simplified related code too.

dimiii:
@mberlanda Frankly, I have lost interest. Can you close the MR, please?

Issue #137, opened 2017-02-20
does sidekiq-cron know how to handle multiple sidekiq processes and not mess up enqueuing? 


dbenjamin57:

ondrejbartas:
Hi sidekiq cron can be used for applications which runs on multiple nodes or have more sideqk processes running.

The whole adding system is based on sorted set of redis which is used to determine that only first process who ask can enque job into queue. Any other process will as same sorted set and will get false, so they know that job was already enqueud for that time. 

Be aware that time is important for sidekiq cron so all servers should be synced with some global time provider. Then you will not face some problem which can occure (but this problem is over all for sidekiq, database etc...)

PR #176, opened 2017-10-03
README is explicit that it's safe to use sidekiq-cron with multiple processes

joshuacronemeyer:
Hi,

Thanks for this gem. I spent a little bit of time today worrying about the possibility of sidekiq-cron running duplicate jobs in heroku when I scale up the number of sidekiq worker dynos. Came across https://github.com/ondrejbartas/sidekiq-cron/issues/137 which made it clear I didn't need to worry about that.

Thought it would be nice to mention this in the README. I pretty much just recycled @ondrejbartas explanation from the referenced issue. Hopefully it's accurate.


ondrejbartas:
@joshuacronemeyer thanks for that

PR #188, opened 2018-02-10
Fix poll interval description in README

tisba:
I noticed that the README is outdated :)

The default poll interval is now 30s not 10s.


ondrejbartas:
@tisba thank you for update

PR #194, opened 2018-03-22
Add 'wrapped' argument to ActiveJob message

matsimitsu:
Fixes issue where sidekiq-cron doesn't add the "wrapped" argument that is used by ActiveJob and Sidekiq to determine the original class

Rails has a test for it: https://github.com/rails/rails/blob/9d9f752661c31b3063d55bec14e797c957d2bb7d/activejob/test/integration/queuing_test.rb#L30 

And Sidekiq uses it in their API: https://github.com/mperham/sidekiq/blob/2ed92600fa71a9c275189d01df369ad4f8b9ca32/lib/sidekiq/api.rb#L322

Without the 'wrapped' param it takes the entire "args" class as the job name, causing all kinds of issues.

### DIFF
--- a/lib/sidekiq/cron/job.rb
+++ b/lib/sidekiq/cron/job.rb
@@ -124,6 +124,7 @@ module Sidekiq
       def active_job_message
         {
           'class'        => 'ActiveJob::QueueAdapters::SidekiqAdapter::JobWrapper',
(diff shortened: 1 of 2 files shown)


ondrejbartas:
@matsimitsu thanks for PR

PR #204, opened 2018-06-15
Update rufus scheduler dependency

basmoura:
**Problem**

The last version of [rufus-scheduler](https://github.com/jmettraux/rufus-scheduler/releases/tag/v3.5.0) brakes the job creation on sidekiq-cron, perhaps due the implementation on Fugit gem on their side.

**Solution**
To maintain the regular behavior  to create the jobs, I've defined the version `~> 3.4.0` of the `rufus-scheduler` until they fix this on their side. (which i believe it will take some time).


jmettraux:
@basmoura wrote:
> To maintain the regular behavior to create the jobs, I've defined the version ~> 3.4.0 of the rufus-scheduler until they fix this on their side. (which i believe it will take some time)

@basmoura, what makes you believe it will take some time to the rufus-scheduler authors to fix this on their side? What makes you believe the fix has to happen on their side?

PR #210, opened 2018-07-16
Switch Time.strptime to DateTime.strptime to always parse as UTC

film42:
This is alternate implementation of https://github.com/ondrejbartas/sidekiq-cron/pull/209, although I do believe https://github.com/ondrejbartas/sidekiq-cron/pull/209 to be the more correct implementation. Only difference is that this one doesn't require a breaking change.

As far as I can tell `Time.strptime` will use the system time, `DateTime.strptime` will assume the date string is UTC.

Example:
```ruby
irb(main):004:0> LAST_ENQUEUE_TIME_FORMAT = '%Y-%m-%d %H:%M:%S'
=> "%Y-%m-%d %H:%M:%S"
irb(main):005:0> Time.now
=> 2018-07-16 16:30:47 -0400
irb(main):029:0> ref_time = Time.now.utc
=> 2018-07-16 20:33:44 UTC
irb(main):030:0> time_string = ref_time.strftime(LAST_ENQUEUE_TIME_FORMAT)
=> "2018-07-16 20:33:44"
irb(main):031:0> Time.strptime(time_string, LAST_ENQUEUE_TIME_FORMAT)
=> 2018-07-16 20:33:44 -0400
irb(main):032:0> DateTime.strptime(time_string, LAST_ENQUEUE_TIME_FORMAT)
=> Mon, 16 Jul 2018 20:33:44 +0000
irb(main):033:0> DateTime.strptime(time_string, LAST_ENQUEUE_TIME_FORMAT).to_i
=> 1531773224
irb(main):034:0> Time.strptime(time_string, LAST_ENQUEUE_TIME_FORMAT).to_i
=> 1531787624
irb(main):035:0> ref_time.to_i
=> 1531773224

```

PR #209, opened 2018-07-13
Corrected bug on @last_enqueue_time

Akhilesh05:
I enqueued a cron, instantly the **Last enqueued** column was 4 hours ago.
I am at +4 GMT. I found that the `@last_enqueued_time` is added in redis using:
```ruby
LAST_ENQUEUE_TIME_FORMAT = '%Y-%m-%d %H:%M:%S'
```
It indeed lacks the timezone.
See this:
```ruby
LAST_ENQUEUE_TIME_FORMAT = '%Y-%m-%d %H:%M:%S'
=> "%Y-%m-%d %H:%M:%S"
time = Time.now.utc
=> 2018-07-13 16:50:26 UTC
str_time = time.strftime(LAST_ENQUEUE_TIME_FORMAT)
=> "2018-07-13 16:50:26"
Time.strptime(str_time, LAST_ENQUEUE_TIME_FORMAT).utc
=> 2018-07-13 12:50:26 UTC # BUG
```
By just adding the timezone with `'%Y-%m-%d %H:%M:%S %z'`, I fixed it.
### See bug:
bug image

PR #213, opened 2018-07-18
Support timezones in version 1.0.0

peterzhu2118:
A had a problem running sidekiq-cron on my development machine because I was located in EST time zone, while the sidekiq-cron stored my time as UTC but parsed it as EST, causing a "Last enqueued" date in the future, like this:
image

This meant that the cron job that should have been running once every minute was actually running once every 4 hours.

Before the change:
```ruby
[17] pry(main)> Sidekiq::Cron::Job.all
=> [#<Sidekiq::Cron::Job:0x00007feb6e3f1d60
  @active_job=false,
  @active_job_queue_name_delimiter="",
  @active_job_queue_name_prefix="",
  @args=[],
  @cron="* * * * *",
  @description="",
  @fetch_missing_args=false,
  @klass="GithubTokenRefresherJob",
  @last_enqueue_time=2018-07-18 14:38:32 -0400,
  @message="{\"retry\":true,\"queue\":\"default\",\"class\":\"GithubTokenRefresherJob\",\"args\":[]}",
  @name="github_token_refresher_job",
  @queue="default",
  @queue_name_with_prefix="default",
  @status="enabled">]
```
Notice that `last_enqueue_time` contained wrong timezone info, it was 10:38AM at EST timezone, which is 14:38 in UTC, but it stores the timezone as EST, which is the problem.

After the change:
```ruby
[6] pry(main)> Sidekiq::Cron::Job.all
=> [#<Sidekiq::Cron::Job:0x00007f8a30e76410
  @active_job=false,
  @active_job_queue_name_delimiter="",
  @active_job_queue_name_prefix="",
  @args=[],
  @cron="* * * * *",
  @description="",
  @fetch_missing_args=false,
  @klass="GithubTokenRefresherJob",
  @last_enqueue_time=2018-07-18 14:40:22 +0000,
  @message="{\"retry\":true,\"queue\":\"default\",\"class\":\"GithubTokenRefresherJob\",\"args\":[]}",
  @name="github_token_refresher_job",
  @queue="default",
  @queue_name_with_prefix="default",
  @status="enabled">]
```
Notice that `last_enqueue_time` now has the correct timezone. I ran this in my local environment, and it seems to work as intended.

### DIFF
--- a/lib/sidekiq/cron/job.rb
+++ b/lib/sidekiq/cron/job.rb
@@ -12,7 +12,7 @@ module Sidekiq
 
       #how long we would like to store informations about previous enqueues
       REMEMBER_THRESHOLD = 24 * 60 * 60
-      LAST_ENQUEUE_TIME_FORMAT = '%Y-%m-%d %H:%M:%S'
+      LAST_ENQUEUE_TIME_FORMAT = '%Y-%m-%d %H:%M:%S %z'
 
       #crucial part of whole enquing job
       def should_enque? time
--- a/test/unit/job_test.rb
+++ b/test/unit/job_test.rb
@@ -600,7 +600,7 @@ describe "Cron Job" do
 
     it "last_enqueue_time shouldn't be rewritten after save" do
       #adding last_enqueue_time to initialize is only for test purpose
-      last_enqueue_time = '2013-01-01 23:59:59'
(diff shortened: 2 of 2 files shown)


jmettraux:
gh-209 and gh-210

Issue #220, opened 2018-09-04
retrieve the scheduled date of a sidekiq cron worker within it


ngouy:
Out of the box, my point is the following : Is there a way (without passing it as a param) to retrieve the original scheduled_date of a worker ? regardless of the retries number, or the queue delaying ?

Is it possible to think of an additional param to the perform method, corresponding of the original scheduled_at ?

ngouy:
#221

PR #221, opened 2018-09-05
add date_as_argument job new property

ngouy:
new mandatory job property to have the original execution date of the job as last argument of the perform method.

Useful if you need to use the theoretical execution time into your worker but is delayed due to insufficient threads number, or retries.


ngouy:
issue #220

ondrejbartas:
Hi released in `1.1.0`

PR #225, opened 2018-10-01
Updated README to use current version

corroded:
I think it's okay to use the current version here because we already have the upgrade warning in place any way.


ondrejbartas:
Hi released in `1.1.0`

PR #232, opened 2018-12-01
Update README in respect of rufus-scheduler bug

adrianobarroso:
I encountered a problem in production that persisted for quite time until I found out that sidekiq-cron versions lower than `1.0` have a problem with rufus-scheduler version `3.5`.

Maybe updating the README with that note will be useful for other people. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,8 @@ Upgrade from <0.6x to 1.0.x
 
 Be aware that there was change from rufus-scheduler to fugit. Also there was change in handling Timezones, so after upgrade check that your jobs are enqueued in correct times. Best thing to do is correctly set timezone in cron notations.
 
+The issue is versions `3.5` of rufus-scheduler are not compactible with versions under `1.0` of this gem, be aware that `Sidekiq::Cron::Job.create` will return false and not schedule your job/worker.
+
 Requirements
 -----------------
 


jmettraux:
Good morning,

may I suggest the following text instead:

> Please be aware that sidekiq-cron < 1.0 was relying on rufus-scheduler < 3.5. Using those older sidekiq-cron with rufus-scheduler >= 3.5 ends up with jobs failing on creation. Sidekiq-cron 1.0 includes a patch that switches from rufus-scheduler to rufus-scheduler's core dependency, fugit.

I hate to read that rufus-scheduler is not "compactible" with a project that had an open dependency (>= 3.3) on it.

The patch: https://github.com/ondrejbartas/sidekiq-cron/pull/200
Learnings: http://jmettraux.skepti.ch/20180601.html?t=Gemfile_lock
The uncompactibles: https://github.com/jmettraux/rufus-scheduler and https://github.com/floraison/fugit

Best regards.

adrianobarroso:
Sure @jmettraux, much better and clean explanation. Its done [here](https://github.com/ondrejbartas/sidekiq-cron/pull/232/commits/c352ce24c963a39b15f5e76fa0af6d38a1150a20).

Think this README improve its better for people facing this problem to catch it quickly. :)

ondrejbartas:
Hi released in `1.1.0`

PR #252, opened 2019-06-16
Clarify in readme where to call create

collimarco:
You must call Sidekiq::Cron::Job.create in a file different from the cron job.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,11 +115,14 @@ class HardWorker
   end
 end
 
+# then in another file...
 Sidekiq::Cron::Job.create(name: 'Hard worker - every 5min', cron: '*/5 * * * *', class: 'HardWorker') # execute at every 5 minutes, ex: 12:05, 12:10, 12:15...etc
 # => true
 ```
 
 `create` method will return only true/false if job was saved or not.
+
+You must call `create` in a file different from the cron job, otherwise the file will never be executed due to Rails lazy loading.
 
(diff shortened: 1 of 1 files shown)


ondrejbartas:
@collimarco I will get executed when you are running in production. But best place to put create is in `cofing/initializers/...`

collimarco:
Keep open

markets:
So what's missing here @ondrejbartas @collimarco? I've never used this approach when using `sidekiq-cron` so I have no strong opinion about this one 🙏🏼 

@ondrejbartas proposed here https://github.com/ondrejbartas/sidekiq-cron/pull/252#issuecomment-608428282 to recommend people to put them under an initializer.

honzasterba:
I don't like the approach because it is specific to Rails. What about adding a Troubleshooting section to the readme. Or just another section around [here](https://github.com/ondrejbartas/sidekiq-cron#forking-processes-or-problem-with-notimplementederror).

markets:
Hello @collimarco 👋🏼 

I agree with @honzasterba, this is too much Rails specific... I propose to create a new section, called "Troubleshooting" and add there the current note about [forking processes](https://github.com/ondrejbartas/sidekiq-cron#forking-processes-or-problem-with-notimplementederror) and this new one about Rails lazy load.

collimarco:
Keep open

PR #268, opened 2020-01-17
Implements namespaces

zedtux:
#  Namespacing

This gem can isolate jobs within namespaces.

## Default namespace

When not giving a namespace, the `default` one will be used.

In the case you'd like to change this value, create a new initializer like so:

`config/initializers/sidekiq-cron.rb`:

```ruby
  config.default_namespace = 'statics'
end
```

## Usage

When creating a new job, you can optionaly give a `namespace` attribute, and then you can pass it too in the `find` or `destroy` methods.

```ruby
Sidekiq::Cron::Job.create(
  name: 'Hard worker - every 5min',
  namespace: 'Foo',
  cron: '*/5 * * * *',
  class: 'HardWorker'
)
# INFO: Cron Jobs - add job with name Hard worker - every 5min in the namespace Foo

# Without specifing the namespace, Sidekiq::Cron use the `default` one, therefore `count` return 0.
Sidekiq::Cron::Job.count
#=> 0

# Searching in the job's namespace returns 1.
Sidekiq::Cron::Job.count 'Foo'
#=> 1

# Same applies to `all`. Without a namespace, no jobs found.
Sidekiq::Cron::Job.all

# But giving the job's namespace returns it.
Sidekiq::Cron::Job.all 'Foo'
#=> [#<Sidekiq::Cron::Job:0x00007f7848a326a0 ... @name="Hard worker - every 5min", @namespace="Foo", @cron="*/5 * * * *", @klass="HardWorker", @status="enabled" ... >]

# If you'd like to get all the jobs across all the namespaces then pass an asterisk:
Sidekiq::Cron::Job.all '*'
#=> [#<Sidekiq::Cron::Job ...>]

job = Sidekiq::Cron::Job.find('Hard worker - every 5min', 'Foo').first
job.destroy
# INFO: Cron Jobs - deleted job with name Hard worker - every 5min from namespace Foo
#=> true
```

## Sidekiq web update

A new bar has been added showing the namespaces, allowing to switch namespaces:

Screenshot 2020-01-21 at 09 19 04

# Notes

I haven't updated the slim files as slim has been removed 7 years ago from sidekiq, but if you really want it, I can update them.


ondrejbartas:
@zedtux Very nice code. I like your approach. Can you rebase your code to master and change slim templates? I will like to release it with new version

zedtux:
Thank you @ondrejbartas. Sure I'll do it.

For your information, this code is running in our prod since weeks.

zedtux:
@ondrejbartas it's done !

zedtux:
Don't know what is happening ... Github seem to have missed the Travis CI callback to set the build as passed 🤷‍♂.

zedtux:
@ondrejbartas can you please have a look?

zedtux:
@ondrejbartas please?

zedtux:
@ondrejbartas is there any chance this get merged one day or should I wast my time in building a new gem?

honzasterba:
@zedtux please update the PR to latest master, would love to merge this

zedtux:
@honzasterba all conflicts fixed, can you please merge it now ?

README.md
```diff
@@ -60,6 +60,7 @@ gem "sidekiq-cron", "~> 1.3"
   'cron' => '1 * * * *',  # execute at 1 minute of every hour, ex: 12:01, 13:01, 14:01, 15:01, ... (HH:MM)
   'class' => 'MyClass',
   # OPTIONAL
+  'namespace' => 'YourNamespace', # groups jobs together in a namspace (Default value is 'default'),
```
honzasterba, line 63: type in `namspace`

README.md
```diff
@@ -82,6 +83,60 @@ If using Rails, this is evaluated against the timezone configured in Rails, othe
...
+
+# Same applies to `all`. Without a namespace, no jobs found.
+Sidekiq::Cron::Job.all
+
+# But giving the job's namespace returns it.
+Sidekiq::Cron::Job.all 'Foo'
+#=> [#<Sidekiq::Cron::Job:0x00007f7848a326a0 ... @name="Hard worker - every 5min", @namespace="Foo", @cron="*/5 * * * *", @klass="HardWorker", @status="enabled" ... >]
+
+# If you'd like to get all the jobs across all the namespaces then pass an asterisk:
+Sidekiq::Cron::Job.all '*'
```
honzasterba, line 131: I am not sure about this API. Why does `Sidekiq::Cron::Job.all` not return all jobs in all namespaces?

Basically it would seem more logical to me that not specifying a namespace would search all jobs.

lib/sidekiq/cron.rb
```diff
@@ -1,8 +1,30 @@
 require "sidekiq/cron/job"
+require "sidekiq/cron/namespace"
 require "sidekiq/cron/poller"
 require "sidekiq/cron/launcher"
 
 module Sidekiq
   module Cron
+    Redis.respond_to?(:exists_returns_integer) && Redis.exists_returns_integer =  false
```
honzasterba, line 8: this is deprecated and should be removed

lib/sidekiq/cron/job.rb
```diff
@@ -202,13 +204,20 @@ def self.load_from_array! array
       end
 
       # get all cron jobs
-      def self.all
+      def self.all(namespace = Sidekiq::Cron.configuration.default_namespace)
```
honzasterba, line 207: if this were `self.all(namespace = nil)` and would mean `ignore namespaces` it would seem more logical to me

honzasterba (changes requested):
missin unit tests

README.md
zedtux: I made `Sidekiq::Cron::Job.all` returning the jobs from the `default_namespace` allowing one to fetch jobs only from that namespace (like Kubernetes does for example), and when you want really ALL the jobs, then you pass the asterisk symbol.

That look logical to me, but if you really want `Sidekiq::Cron::Job.all` to return ALL jobs by default, then you need to clarify when is the `default_namespace` used by the user please.

lib/sidekiq/cron.rb
zedtux: Sorry about this one, I was confused during the `rebase`.

lib/sidekiq/cron/job.rb
zedtux: I understand you point, but I then don't see how users will use the `default_namespace`.

zedtux:
Unit tests added

lib/sidekiq/cron/web_extension.rb
```diff
@@ -3,70 +3,112 @@ module Cron
...
+          @namespaces = Sidekiq::Cron::Namespace.all_with_count
+
+          @cron_jobs = Sidekiq::Cron::Job.all(@current_namespace)
+
+          #if Slim renderer exists and sidekiq has layout.slim in views
+          if defined?(Slim) && File.exists?(File.join(settings.views,"layout.slim"))
+            render(:slim, File.read(File.join(view_path, "cron.slim")))
+          else
+            render(:erb, File.read(File.join(view_path, "cron.erb")))
+          end
```
markets, line 36: This block can be simplified as:

```suggestion
          render(:erb, File.read(File.join(view_path, "cron.erb")))
```

README.md
honzasterba: I am not sure I understand your comment.

`Sidekiq::Cron::Job` methods `all`, `count` and others should return all jobs regardless of namespace
user can call `Sidekiq::Cron::Job.all('default')` to get the jobs from the default namespace

zedtux: The goal of namespaces here is to segregate jobs in groups so that one can execute actions on all the jobs from a specific namespace, and be sure that other jobs from other namespaces are untouched.

Forcing the dev to tell the namespace from which to execute that action, or forcing the dev to retrieve all the jobs by passing an asterisk prevent him from doing a mistake and ending by deactivating, or worst destroying all the jobs from all the namespaces.

At the end of the day, for all the devs using this gem, when they don't want to use namespaces, they just have to ignore it and the gem will still work in the same way : all the jobs will go to the default namespace specified by the `default_namespace` configuration option (which is `default` by default), and using the `Sidekiq::Cron::Job` methods will return all their jobs without passing any argument (see all your tests are still passing).

From our point of view, this is a nice trade-off :
* people which doesn't care about namespaces will use exactly the same way this gem
* people using namespaces will tell what they want (a specific namespace, or no namespaces and get all the jobs across all the existing namespaces) so that they can't make mistakes.

honzasterba: I agree with you, but: 
What concerns me is that Job.all(nil) has undefined behaviour in the current code. 
My thinking is that you either need to add a check for nil into jobs_key() and raise an error when nil is passed
or make nil represent the "*". Which to me looks like to cleaner option because it means less code (no special handling for nil and "*", no method arg default value reading from config) and that means less complexity a cleaner API. 
I think that nobody will be surprised that Job.all will return all jobs from all namespaces.
Another solution that comes to mind is to remove this arg from Job completely and have it always return everything (just like before this PR). And use a different API to work with namespaced Jobs like `Sidekiq::Cron::Namespace.new('default').jobs`

zedtux: I see, that's a good catch. I just pushed a fix for that case with more unit test testing with `nil`, `""` and when namespace doesn't exist.

Could you please have a look and give us your feedback ?

honzasterba: Thanks for the quick fix, however I don't agree with the direction you are taking. 
@markets WDYT?

markets: My team does not plan to use `namespaces` for now, so I 100% agree with that point:

> At the end of the day, for all the devs using this gem, when they don't want to use namespaces, they just have to ignore it and the gem will still work in the same way

So I think the trade-off presented by @zedtux makes sense: for people not using `namespaces` this should be a "transparent" upgrade with no risk, and then if someone wants/needs to use `namespaces` they should pass the namespace in all operations.

honzasterba: @markets yes, we are all in agreement in that, the question is about the API design for getting all jobs.

`Sidekiq::Cron::Job.all('*')`  vs `Sidekiq::Cron::Job.all`

markets: What about the following API?
- `Sidekiq::Cron::Job.all('*')` => returns all jobs from all namespaces
- `Sidekiq::Cron::Job.all` => returns all jobs from the `default_namespace`
- `Sidekiq::Cron::Job.all(namespace)` => returns all jobs from the given `namespace`

That way, for people not using `namespaces` functionality, they can continue using same API with no changes (`Sidekiq::Cron::Job.all`).

honzasterba: that is how it is implemented today, hence I will close do a final review and merge

lib/sidekiq/cron/job.rb
```diff
@@ -618,45 +639,53 @@ def parse_enqueue_time(timestamp)
...
-          # or could it be?
-        #last_cron_time = last_time(current_time)
+        # or could it be?
+        # last_cron_time = last_time(current_time)
         return false if (current_time.to_i - last_cron_time.to_i) > 60
         true
       end
 
+      def self.default_if_blank(namespace)
+        if namespace.nil? || namespace == ''
```
honzasterba, line 649: what about

```
if namespace.nil? || namespace == ''
    Sidekiq::Cron.configuration.default_namespace
else
    namespace
end
```

lib/sidekiq/cron/namespace.rb
```diff
@@ -0,0 +1,54 @@
+require 'sidekiq'
+require 'sidekiq/util'
+
+# Sidekiq::Cron::Namespace
```
honzasterba, line 4: not sure what is the point of this coment

lib/sidekiq/cron/namespace.rb
```diff
@@ -0,0 +1,54 @@
...
+module Sidekiq
+  module Cron
+    class Namespace
```
honzasterba, line 7: missing unit tests for this class

honzasterba (changes requested):

README.md
zedtux: One last point coming to my mind, is "migration". I will test the namespace code with a job without a namespace in order to avoid breaking existing jobs when migrating to the newer version of the gem.

markets: Yep, we should ensure "migration" is gonna be smooth for all current installations and people can just upgrade the gem and everything is working as usual.

zedtux:
I've updated the code in order to comply with Honza's comments.

I'm now working on dealing with the jobs, created prior to this PR, meaning without a defined namespace. There are two options in my eyes:
* having a task to migrate the jobs (copying the jobs to the right place and delete them)
* having the gem checking all the time for jobs without a defined namespace and 
  * migrate them when working on them
  * work with them without migrating them at all

The later would be more resources consumption but transparent, while the first one is the complete opposite.

What would you like me to do here ?

markets:
> - having a task to migrate the jobs

That option seems probably too much breaking from my pov. I'd try to avoid breaking changes, even in a future v2. This kind of changes make the upgrades harder (quite sure some apps will stick to v1, other will miss jobs, ...). 

So, the "transparent" option seems definitely more smooth and the path we should try to implement. But to be honest, not sure if those changes make the code complicated and hard to maintain.

zedtux:
Thank you for your input, I will go for that option.

zedtux:
Yep, I'm busy right now but this PR is in my priority list

zedtux:
Our big migration is done, I will have time now to work on this one.

zedtux:
Back on it ! :muscle:

lib/sidekiq/cron/job.rb
```diff
@@ -638,41 +662,82 @@ def not_past_scheduled_time?(current_time)
...
+      def self.migrate_old_jobs_if_needed!
+        Sidekiq.redis do |conn|
+          old_job_keys = conn.smembers('cron_jobs')
+          old_job_keys.each do |old_job|
+            old_job_hash = conn.hgetall(old_job)
+            old_job_hash[:namespace] = Sidekiq::Cron.configuration.default_namespace
+            create(old_job_hash)
+          end
+
+          old_job_keys.each { |old_job_key| conn.srem('cron_jobs', old_job_key) }
```
honzasterba, line 698: why do we need to loop the old_job_keys again here? why not just call the `conn.srem` in the first loop after migrated job has been created?

README.md
honzasterba: Ok, one final argument and then I will give up on this debate:
Currently my code uses Sidekiq::Cron::Job.all to do something all the Cron jobs scheduled. Once I switch to namespaces for some component of mine this call will suddenly change behaviour because it will no longer return all the jobs butly only the ones in `default` namespace. This seems like a unnecessary breaking change.
What do you think @markets ?

markets: @honzasterba @zedtux Even I like the concept of having the ability to define "namespaces", I'm starting to think that this change will be hard to release without any kind of breaking change... Maybe we should release it as a separate plugin for now? Something like `sidekiq-cron-namespaces`... So people can decide if they want to use the feature and "fight" with potential problems during the migration process.

As I said before, I like this idea... but being honest, the migration seems a bit "dangerous" and there is a lot of changes (in critical paths) to review in that branch that I'm not confident about...

lib/sidekiq/cron/namespace.rb
```diff
@@ -0,0 +1,45 @@
...
+          }
+        end
+      end
+
+      def self.count(name = Sidekiq::Cron.configuration.default_namespace)
+        out = 0
+        Sidekiq.redis do |conn|
+          out = conn.scard("cron_jobs:#{name}")
+        end
+        out
```
serprex, line 41:
```suggestion
        Sidekiq.redis do |conn|
          conn.scard("cron_jobs:#{name}")
        end
```

zedtux:
I don't know why but I can reply you @honzasterba to your last comment, so I'm doing it here :

> Currently my code uses Sidekiq::Cron::Job.all to do something all the Cron jobs scheduled. Once I switch to namespaces for some component of mine this call will suddenly change behaviour because it will no longer return all the jobs butly only the ones in default namespace. This seems like a unnecessary breaking change.

You are missing here the fact that jobs are automagically migrated from the _no namespace space_ to the _default namespace_ which means at then end nothing change for you. Don't look the technical change that this PR brings, but the end user experience instead.
This PR doesn't break anything anymore since I have added that automatic migration code. Or did I miss-understood something from your "final argument" ?

README.md
markets: @zedtux ok! we'll need to perform a more in depth review (sorry 🙏🏼 I didn't touched this part of the codebase and I'd be more confident if other maintainers/contributors also review this branch), but +1 to ship this in future releases. 

I just made a quick re-review and it seems we are on the right path now (especially regarding the auto-migration process). I'll test this out in my apps.

But, I think we should focus first on finish the Sidekiq v7 support, which is happening on #369 and touches also some internals.

Thank you so much for your work!

zedtux: @markets I just realise now you've done the Sidekiq v7 support. The awesome bot is starting to bother on this PR now ... and it's the new year, so what's the plan ? Rebase / retest / merge ? 😎

markets: Hi @zedtux 👋🏼 

Yeah, I think we are ready to continue with this branch. IMO next steps should be: rebase, finish review (I'd like to hear final thoughts from @honzasterba and/or @ondrejbartas too), test and merge 🚀

zedtux: Do you have any news @markets please?

markets: I'm overall ok with the feature, but we still need to resolve some @honzasterba's comments (and update from master, we added quite new things and fixes lately).

zedtux: Okay, I'll go through the comments one more time, hopefully the last time. Thanks.

zedtux:
Yep, I need to find some time to work again on this one.

zedtux:
I have managed the `rebase` with the `master` branch, I'm now looking to the code review comments.

lib/sidekiq/cron/namespace.rb
zedtux: This has been solved in the `master` branch, and the `rebase` propagated it in my branch.

zedtux:
As I have said in #406, my system is container only based, therefore this PR has the `Dockerfile` and `docker-compose.yml` files back.

I have used them to run the tests in order to develop this PR.

In the case you don't like it and would like me to remove them, just ask, but it is really useful to prevent messing a machine.

markets:
@zedtux About Docker, I started a conversation some months ago: https://github.com/sidekiq-cron/sidekiq-cron/discussions/383 with no success.

To re-introduce Docker support, I'd like to make some changes (sorry if I say something crazy, I'm not a Docker user 😅):
- Use a more official image (Alpine Ruby?)
- Move Docker related files to a folder (docker/*) to avoid polluting the root workspace
- Document how to use it (probably under this section https://github.com/sidekiq-cron/sidekiq-cron#contributing)

zedtux:
> Use a more official image (Alpine Ruby?)

I'm not sure what you mean here, `ruby:2.7` refers to the officiel Ruby docker image registry : https://hub.docker.com/_/ruby. The Alpine version is a smaller version as a label from the same repository.

I could update it, if you really have an interest for it, but in my honest opinion, as this is only a development image, it doesn't matter much.\
Image size is more about production environment in order to speed-up the deployment.

> remove MAINTAINER Joao Serra <joaopfserra@gmail.com> (he is not maintaining anything in this organization)

Sure, it has to be changed to something else. Like you ? Or you don't want to specify a `MAINTAINER`?

> Move Docker related files to a folder (docker/*) to avoid polluting the root workspace

I'm used to keep them at the root so that the commands are easier but I can understand your point of view since you're not a docker user.

> Document how to use it (probably under this section https://github.com/sidekiq-cron/sidekiq-cron#contributing)

I will do it with pleasure.

Otherwise, about this PR, can you do the code review and merge or is it still @honzasterba who's doing it?

zedtux:
Okay, so let's remove the `MAINTAINER` instruction, upgrade to ruby 3.2 and move files in a `docker/` folder.

zedtux:
Could anyone do the code review please, I really would love to see this merged! 😍

lib/sidekiq/cron/job.rb
```diff
@@ -464,19 +475,18 @@ def save
...
 
         Sidekiq.redis do |conn|
-
           # Add to set of all jobs
-          conn.sadd self.class.jobs_key, [redis_key]
+          conn.sadd self.class.jobs_key(@namespace), [redis_key]
 
           # Add informations for this job!
-          conn.hset redis_key, to_hash.transform_values! { |v| v || "" }
+          conn.hset redis_key, (to_hash.transform_values! { |v| v || '' })
```
serprex, line 482: This line doesn't need to be changed

lib/sidekiq/cron/job.rb
```diff
@@ -529,9 +539,17 @@ def self.destroy_all!
 
       # Remove "removed jobs" between current jobs and new jobs
       def self.destroy_removed_jobs new_job_names
-        current_job_names = Sidekiq::Cron::Job.all.map(&:name)
+        current_jobs = Sidekiq::Cron::Job.all('*')
```
serprex, line 542: Existing code looks to use double quotes, should maintain consistent quoting style

lib/sidekiq/cron/job.rb
```diff
@@ -631,36 +650,70 @@ def not_past_scheduled_time?(current_time)
...
+        else
+          namespace
+        end
+      end
+
+      def self.job_keys_from_namespace(namespace = Sidekiq::Cron.configuration.default_namespace)
+        Sidekiq.redis do |conn|
+          if namespace == '*'
+            namespaces = conn.keys(jobs_key(namespace))
+            namespaces.collect { |name| conn.smembers(name) }.flatten
```
serprex, line 665:
```suggestion
            namespaces.flat_map { |name| conn.smembers(name) }
```

lib/sidekiq/cron/job.rb
zedtux: Well that line is marked with the following from Rubocop:

> Lint/AmbiguousBlockAssociation - Parenthesize the param `to_hash.transform_values! { |v| v || '' }` to make sure that the block will be associated with the `to_hash.transform_values!` method call.

Whick looked important enough to me to change it, but I will revert it.

zedtux:
> Okay, so let's remove the `MAINTAINER` instruction, upgrade to ruby 3.2 and move files in a `docker/` folder.

This has been done too!

zedtux:
I don't understand why build matrix is red now, it doesn't look related to my changes... anyone has an idea please?

serprex:
@zedtux CI fixed, try updating PR with latest

zedtux:
Thanks @serprex, the build is now green. 👍

zedtux:
@honzasterba @markets the PR is now ready, any other review or are we good to go?

serprex (approved):

markets:
Seems fine to me @zedtux 👍🏼 I'll merge it next week and it will be released as v2.0 soon. Sorry I'm on parental leave and my free time is limited until october.

zedtux:
Hello there,

Any news please ?

markets:
Hi @zedtux, sorry the past 2 weeks we have been working in an relevant fix (more info #428) introduced some months ago.

I think we are now ready to merge this one, targeting a future v2 release.

Let me ask a couple of details 🙏🏼:
- Document how to use `docker` for development (probably around this section https://github.com/sidekiq-cron/sidekiq-cron?tab=readme-ov-file#testing)
- Update screenshot that appears in the README: https://github.com/sidekiq-cron/sidekiq-cron/blob/master/docs/images/web-cron-ui.jpeg

zedtux:
> Let me ask a couple of details 🙏🏼:
> 
> * Document how to use `docker` for development (probably around this section https://github.com/sidekiq-cron/sidekiq-cron?tab=readme-ov-file#testing)
> * Update screenshot that appears in the README: https://github.com/sidekiq-cron/sidekiq-cron/blob/master/docs/images/web-cron-ui.jpeg

Hey @markets,

1. Git rebase done
2. Docker doc written
3. Screenshot updated

All should be alright now. Let's merge it!

markets:
Thanks again @zedtux 👍

I'd like to merge it during the christmas holidays and probably cut a release as v2.0.0-beta.

markets (approved):

zedtux:
@markets happy new year, let's get this PR merged! 🥳

PR #281, opened 2020-04-07
Add rubocop

julienemo:
:heavy_plus_sign: rubocop gem
:heavy_plus_sign: rubocop and test in CI
:wrench: fixed SOME OF the offenses, disabled some

:information_source: 
Still some offenses in `.rubocop_todo.yml`, mainly metrics


markets:
Thanks a lot for this work and sorry that late reply 🙏🏼  

After some time of inactivity, we are doing some issue/pr gardening these days in order to keep this _relevant_ project up to date. I'm going to close this one for now, there are a lot of conflicts (even deleted files) and it will be quite hard to rebase it. 

On the other hand, we are not sure if we want a linter yet. Maybe we'll revisit this topic in the future. We can also analyze other options like [standard](https://github.com/testdouble/standard) and integration in our CI.

Thanks again!

PR #287, opened 2020-06-18
validate jobs in Job.find

dwarburt:
In the case of two or more Sidekiq workers the job hash in Redis can be
left in a partly deleted state after one worker deletes the job and
another worker sets the last_enqueue_time value. If this happens
before this commit then calls to Job.find will return an invalid job
that doesn't have a  name (or any properties but last_enqueue_time)
and can't be destroyed. After this commit nil is returned if the Job is
not valid.

### DIFF
--- a/lib/sidekiq/cron/job.rb
+++ b/lib/sidekiq/cron/job.rb
@@ -233,7 +233,7 @@ module Sidekiq
             output = Job.new conn.hgetall( redis_key(name) )
           end
         end
-        output if output.valid?
+        output if output&.valid?
       end
 
       # create new instance of cron job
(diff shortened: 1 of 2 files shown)


lib/sidekiq/cron/job.rb
```diff
@@ -233,7 +233,7 @@ def self.find name
             output = Job.new conn.hgetall( redis_key(name) )
           end
         end
-        output
+        output if output.valid?
```
dwarburt, line 236:
```suggestion
        output if output&.valid?
```

lib/sidekiq/cron/job.rb
```diff
@@ -233,7 +233,7 @@ def self.find name
             output = Job.new conn.hgetall( redis_key(name) )
           end
         end
-        output
+        output if output&.valid?
```
dwarburt, line 236: Needs ruby 2.2 support.

```suggestion
        output if output && output.valid?
```

ondrejbartas (approved):

PR #290, opened 2020-08-05
Document natural language cron expressions

jeromedalbert:
Fugit understands some kind of [natural language](https://github.com/floraison/fugit#fugitnat).

I have tried it in sidekiq-cron and it works, which is nice when listing scheduled jobs in YML, similarly to the [whenever](https://github.com/javan/whenever) gem.

This PR documents this use case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,6 +197,12 @@ second_job:
   queue: hard_worker_long
   args:
     hard: "stuff"
+
+third_job:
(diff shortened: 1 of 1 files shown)


jeromedalbert:
The natural language syntax actually doesn't currently work with sidekiq-cron. I thought it was working because of previously successfully parsed jobs on my local.

Closing this PR for now, in favor of https://github.com/ondrejbartas/sidekiq-cron/pull/293

PR #296, opened 2021-01-13
Use Redis#exists? instead of Redis#exists

spacebat:
Follow the advice warned by recent versions of redis-rb:

`Redis#exists(key)` will return an Integer in redis-rb 4.3. `exists?` returns a boolean, you should use it instead. To opt-in to the new behavior now you can set Redis.exists_returns_integer =  true. To disable this message and keep the current (boolean) behaviour of 'exists' you can set `Redis.exists_returns_integer = false`, but this option will be removed in 5.0. (~/.rvm/gems/ruby-2.6.3/gems/sidekiq-cron-1.2.0/lib/sidekiq/cron/job.rb:464:in `block in save')

### DIFF
--- a/lib/sidekiq/cron/job.rb
+++ b/lib/sidekiq/cron/job.rb
@@ -461,7 +461,7 @@ module Sidekiq
 
           #add information about last time! - don't enque right after scheduler poller starts!
           time = Time.now.utc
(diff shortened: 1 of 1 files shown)


spacebat:
D'oh leaving this to https://github.com/ondrejbartas/sidekiq-cron/pull/288

PR #297, opened 2021-02-05
Fix redis 4.2.0 Redis#exists warning

itsgg:
Fixes the following deprecation warning

```
`Redis#exists(key)` will return an Integer in redis-rb 4.3. `exists?` returns a boolean, you should use it instead. To opt-in to the new behavior now you can set Redis.exists_returns_integer =  true. 
To disable this message and keep the current (boolean) behaviour of 'exists' you can set `Redis.exists_returns_integer = false`, but this option will be removed in 5.0. (/Users/gg/.asdf/installs/ruby/2.6.6/lib/ruby/gems/2.6.0/gems/redis-namespace-1.8.0/lib/redis/namespace.rb:476:in `call_with_namespace')
```

### DIFF
--- a/lib/sidekiq/cron/job.rb
+++ b/lib/sidekiq/cron/job.rb
@@ -461,7 +461,7 @@ module Sidekiq
 
           #add information about last time! - don't enque right after scheduler poller starts!
           time = Time.now.utc
-          conn.zadd(job_enqueued_key, time.to_f.to_s, formated_last_time(time).to_s) unless conn.exists(job_enqueued_key)
+          conn.zadd(job_enqueued_key, time.to_f.to_s, formated_last_time(time).to_s) unless conn.exists?(job_enqueued_key)
         end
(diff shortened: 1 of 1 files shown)


MikeRogers0:
A PR exists for this https://github.com/ondrejbartas/sidekiq-cron/pull/288 ;)

honzasterba:
@ondrejbartas I offer my help maintaining this gem, I was about to create the same PR but found this being here un-merged for quite some time

MikeRogers0:
@ondrejbartas I'd love to assist with maintenance also! I'm a massive fan of this gem & I'd be happy to pay for your time if you want to look at this :)

CC: @honzasterba

ondrejbartas (approved):

honzasterba:
we also need to update the depency version for redis-namespace

markets:
Closing as #288 was already merged, in case we also need to update `redis-namespace`, better to create a new PR. Thank you all!

PR #305, opened 2021-06-09
Make log message more obvious

arrrnas:
`Cron Jobs - add job with name:` seemed like a warning that some action needs to be taken ¯\\_(ツ)\_/¯


Wittiest (approved):

TALlama (approved):
@ondrejbartas This PR matches the "deleted job" message on the other side. Looks good to me.

mathieujobin (changes requested):
i don't see a benefit outside potentially breaking parsing

markets:
@ondrejbartas @honzasterba what do you think about this one? 

I understand @mathieujobin's concern, but consistency is also good!

honzasterba:
Looks good to me.

I don't think log messages should be treated as API.

honzasterba (approved):

PR #311, opened 2022-02-10
fix pipelining redis instance is deprecated

juniortada:
Fix Pipelining commands on a Redis instance is deprecated and will be removed in Redis 5.0.0

```ruby
redis.pipelined do
  redis.get("key")
end

should be replaced by

redis.pipelined do |pipeline|
  pipeline.get("key")
end
```

PR #309, opened 2022-02-04
Fix redis-rb v4.6.0 pipelining depreciation

D1ceWard:
Hi, this is a small fix for the depreciation alert that appeared with version 4.6.0 of redis-rb which can pollute the logs quite a bit.

PR #314, opened 2022-03-10
Migrate CI to GitHub Actions

petergoldstein:
This PR migrates CI to GitHub Actions as Travis CI.org is no longer active.

In addition to adding the GitHub Actions config file, there were a number of other changes made to get the system to green:

1. Modernizing the gemspec - using more modern syntax and conventions in the gemspec and eliminating jeweler
2. Removing various dependencies - there were a number of referenced gems that were not actual development dependencies that made the test suite unrunnable under modern Rubies.  Removing these dependencies addressed this
3. Redis-namespace deprecation warnings - a few changes to address deprecation warnings from `redis-namespace`
4. Changes to the test suite to support CSRF behavior in Sidekiq::Web

Because of these issues this is a bit larger of PR than usual for adding GitHub Actions.  It all runs green on my fork.

I also limited the supported Ruby versions to 2.6-3.1.


petergoldstein:
This addresses #309 , #311 , and #312 as a side effect.

lib/sidekiq/cron/job.rb
```diff
@@ -206,9 +206,9 @@ def self.all
         job_hashes = nil
         Sidekiq.redis do |conn|
           set_members = conn.smembers(jobs_key)
-          job_hashes = conn.pipelined do
+          job_hashes = conn.pipelined do |pipeline|
```
mathieujobin, line 209: I was looking for this !

lib/sidekiq/cron/version.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+module Sidekiq
+  module Cron
+    VERSION = "1.2.0"
```
mathieujobin, line 5: shall we bump it ?

lib/sidekiq/cron/job.rb
petergoldstein: :)

lib/sidekiq/cron/version.rb
petergoldstein: I'm happy to bump it if that's the desire of the maintainers.  I did limit Ruby version to 2.6+, so that may be worth a version bump.

mathieujobin: 1.2.0 is already release and `master` branch already have mentions a 1.3.0 in progress.
I think 1.3.0 would be good.
hopefully someone will release into rubygems

mathieujobin: https://github.com/ondrejbartas/sidekiq-cron/commit/409f11e6eae47df59adcd13cdac4f3c7b5157647

ondrejbartas (approved):

PR #322, opened 2022-04-11
Optimize gem size

markets:
I'd say there is no need to bundle Docker files or screenshots when building the gem. This will reduce the gem size a lot (current gem size: 178 KB, image size: 150K).


honzasterba (approved):
LGTM, thanks!

Issue #247, opened 2019-04-12
`date_as_argument` set as deployment time not time of execution


lloydwatkin:
I'm using YML configuration.

However I see that the cron schedule is stored with the deployment time (or when the sidekiq worker started) as the date argument, whereas I'd expect it to be the time the job was called?

![Screenshot from 2019-04-12 17-23-05](https://user-images.githubusercontent.com/271622/56051787-a7282d00-5d47-11e9-9f50-96cd893dee54.png)

collimarco:
I agree that this is really confusing: I also expected to see the time when the single job instance is created (i.e. a time matching the cron pattern). Unfortunately the time passed as argument is the time when the initializer / sidekiq starts, and it is basically the same for all the job executions.

matiasgarcia:
Just ran into this bug. Agreed that it should be the time at which the job is being added to the queue in sidekiq.

vsuhachev:
try to fix it, #329

markets:
Fixed by #329

PR #329, opened 2022-06-01
fix date_as_argument feature to add timestamp argument at every cron …

vsuhachev:
At Readme `date_as_argument` described as
```
  'date_as_argument' => true, # add the time of execution as last argument of the perform method
```
It is implemented as "add the time of ~~execution~~creation of schedule as last argument of the perform method". It is confusing and mostly useless. If I want to add the creation time of the schedule, then it's easier for me to do it explicitly from the client code.

This PR implements `date_as_argument` feature to add timestamp as extra argument on every job execution

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ vendor/
 Guardfile
 pkg
 .DS_Store
+.idea/
--- a/.gitignore
+++ b/.gitignore
@@ -12,4 +12,3 @@ vendor/
 Guardfile
 pkg
 .DS_Store
-.idea/
--- a/Changes.md
+++ b/Changes.md
@@ -2,13 +2,23 @@
 
 All notable changes to this project will be documented in this file.
 
+## 1.4.0
+
+- Fix buttons order in job show view (https://github.com/ondrejbartas/sidekiq-cron/pull/302)
+- Dark Mode support in UI (https://github.com/ondrejbartas/sidekiq-cron/pull/317/282)
+- Remove invocation of deprecated Redis functionality (https://github.com/ondrejbartas/sidekiq-cron/pull/318)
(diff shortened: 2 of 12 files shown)


markets:
Hi @vsuhachev and thanks for sending this PR! 

It seems ok to me, but since I'm still not familiar with the code base internals, I'd like to wait for feedback from other contributors.

CI is broken due to the Sidekiq upgrade, if you update your branch from current `master`, this will be fixed.

markets:
@ondrejbartas @honzasterba are you agree with this one? Not sure which was the initial idea of that `date_as_argument` attribute

.gitignore
```diff
@@ -12,3 +12,4 @@ vendor/
 Guardfile
 pkg
 .DS_Store
+.idea/
```
honzasterba, line 15: please remove this, this is system specific and should be in your global .gitignore and the change is not related to the PR

honzasterba (commented):
the code looks good to me

the PR description needs to be updated to better document in what situation this behaviour is not working and why this fix is necessary

preferably that would be obvious from the tests, but it is not to me at least

.gitignore
vsuhachev: removed

markets:
@honzasterba Could you please make a second review 🙏🏼?

honzasterba (approved):
LGTM

from the PR description I would think it would be best to remove this feature :)

markets:
Yep, not sure what is the initial idea behind this attribute. We're not using this flag in our projects with `sidekiq-cron` currently, but as per the docs, now it seems "broken" (it does a different thing). 

I'd be ok to completely remove it too 😃, but it seems some people would like it working as described (several +1 in the original issue #247).

honzasterba:
Oh, so people are using it, great lets merge it then!

PR #337, opened 2022-06-10
Adds support for auto-loading the `config/schedule.yml` file

### LINKED ISSUE #250: [Feature] Auto load config/schedule.yml
Creating a `config/schedule.yml` is the most clear way to define all the jobs in a single place, in a declarative way.

The documentation suggests to add the following lines to an initializer:

```
schedule_file = "config/schedule.yml"

if File.exist?(schedule_file) && Sidekiq.server?
  Sidekiq::Cron::Job.load_from_hash YAML.load_file(schedule_file)
end
```

I think that **the above code should be included directly in the gem**: if someone creates a `config/schedule.yml` and installs this gem clearly wants to load that file.

sandip-mane:
fixes #250

With this change, we do not need to explicitly keep an initializer to load the `config/schedule.yml` file. If someone wants to use a different name, that's possible by putting the following configuration in sidekiq options.

```yml
default: &default
  :concurrency: 5

production:
  <<: *default
  cron_schedule_file: "config/schedule_prod.yml"

queues:
  - default
```

For some reason, if someone decides to go the initializer way, that'll work too. Have verified that the jobs don't enqueue again.

As an added benefit, this gives flexibility in having different schedules for different environments. 
_(Not sure why that will be needed but, it's there😃)_

----

### Documentation Screenshot:

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -233,21 +233,13 @@ There are multiple ways to load the jobs from a YAML file
 2. When you want to load jobs from a different filename, mention the filename in sidekiq configuration,
 i.e. `cron_schedule_file: "config/users_schedule.yml"`
 3. Load the file manually as follows
+
 ```ruby
 # config/initializers/sidekiq.rb
-schedule_file = "config/users_schedule.yml"
 
-if File.exist?(schedule_file) && Sidekiq.server?
-  Sidekiq::Cron::Job.load_from_hash YAML.load_file(schedule_file)
-end
-```
-
-From version 3.x it is better not to use separate initializer of schedule instead add `config.on(:startup)` to your Sidekiq configuration:
-
-```ruby
 Sidekiq.configure_server do |config|
   config.on(:startup) do
-    schedule_file = "config/schedule.yml"
+    schedule_file = "config/users_schedule.yml"
 
     if File.exist?(schedule_file)
       Sidekiq::Cron::Job.load_from_hash YAML.load_file(schedule_file)
--- a/README.md
+++ b/README.md
@@ -322,30 +322,6 @@ With this, you will get:
 
 ![Web UI](examples/web-cron-ui.png)
 
-### Forking Processes or problem with `NotImplementedError`
-
-If you're using a forking web server like Unicorn you may run into an issue where the Redis connection is used
-before the process forks, causing the following exception to occur:
-
-```
-Redis::InheritedError: Tried to use a connection from a child process without reconnecting. You need to reconnect to Redis after forking.
-```
-
-To avoid this, wrap your job creation in the call to `Sidekiq.configure_server`:
-
-```ruby
-Sidekiq.configure_server do |config|
-  config.on(:startup) do
-    schedule_file = "config/schedule.yml"
-
-    if File.exist?(schedule_file)
-      Sidekiq::Cron::Job.load_from_hash YAML.load_file(schedule_file)
-    end
-  end
-end
-```
-
-**NOTE** This API is only available in Sidekiq 3.x.
 
 ## Under the hood
 
--- a/README.md
+++ b/README.md
@@ -227,9 +227,15 @@ second_job:
     hard: "stuff"
 ```
 
+There are multiple ways to load the jobs from a YAML file
+
+1. The gem will automatically load the jobs mentioned in `config/schedule.yml` file.
+2. When you want to load jobs from a different filename, mention the filename in sidekiq configuration,
+i.e. `cron_schedule_file: "config/users_schedule.yml"`
+3. Load the file manually as follows
 ```ruby
 # config/initializers/sidekiq.rb
-schedule_file = "config/schedule.yml"
+schedule_file = "config/users_schedule.yml"
 
 if File.exist?(schedule_file) && Sidekiq.server?
   Sidekiq::Cron::Job.load_from_hash YAML.load_file(schedule_file)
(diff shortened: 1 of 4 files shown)


lib/sidekiq/cron/schedule_loader.rb
```diff
@@ -0,0 +1,16 @@
+require "sidekiq"
+require "sidekiq/cron/job"
+
+if Sidekiq.server?
+  Sidekiq.configure_server do |config|
+    schedule_file = config.options[:cron_schedule_file] || "config/schedule.yml"
+
+    if File.exist?(schedule_file)
+      config.on(:startup) do
+        if File.exist?(schedule_file)
```
markets, line 10: do we need this to be checked a couple of times?

markets:
I'm positive with this change, but we should be sure this change doesn't affect people calling the same code from an initializer.

lib/sidekiq/cron/schedule_loader.rb
sandip-mane: True we don't, I missed it during refactoring. Have updated it.

sandip-mane:
> we should ensure this change doesn't affect people calling the same code from an initializer.

Have tested it manually, it does not cause any issues(no duplicate jobs etc.)
**I tried to look for the code that skips enqueuing jobs if those are already present, could not find it.** But duplicate jobs are not being enqueued.

> Not sure if a warning or similar is necessary to ping them to remove it... 

I think we don't need to. Showing warnings would have made sense if it were to cause any issues in the working.

> Probably some documentation updates are enough?

Agree 👍 I will update the document when the code is finalised as we need to mention the keys etc in the doc.

lib/sidekiq/cron/schedule_loader.rb
```diff
@@ -0,0 +1,14 @@
+require "sidekiq"
+require "sidekiq/cron/job"
+
+if Sidekiq.server?
+  Sidekiq.configure_server do |config|
+    schedule_file = config.options[:cron_schedule_file] || "config/schedule.yml"
+
+    if File.exist?(schedule_file)
+      config.on(:startup) do
+        Sidekiq::Cron::Job.load_from_hash YAML.load_file(schedule_file)
```
sandip-mane, line 10: I myself have this question, why not use `!`? 
I believe whenever the schedule file is updated and some jobs are removed, they need to be removed from schedule too. 

```suggestion
        Sidekiq::Cron::Job.load_from_hash! YAML.load_file(schedule_file)
```

sandip-mane:
**Screenshot of README for this section**

image

sandip-mane:
@markets added suggested changes and documentation. Please review.

danilogco (approved):

markets:
Seems fine to me too ✅ 

I have some doubts 🤔 about keeping this section in the docs (since now this will be automatically loaded): https://github.com/ondrejbartas/sidekiq-cron#forking-processes-or-problem-with-notimplementederror

sandip-mane:
Seems like that section won't be needed now as it's done implicitly. Shall I remove it?

sandip-mane:
@markets I have removed the unwanted documentation section from the README. Please review.

sandip-mane:
@markets shall we be using `Sidekiq::Cron::Job.load_from_hash!` i.e. with `!`? Is there any reason against it?

markets:
> @markets shall we be using `Sidekiq::Cron::Job.load_from_hash!` i.e. with `!`? Is there any reason against it?

Since we are promoting this snippet in docs for years, I'll keep the current "no bang" version, to avoid side effects.

README.md
```diff
@@ -227,9 +227,15 @@ second_job:
...
+2. When you want to load jobs from a different filename, mention the filename in sidekiq configuration,
+i.e. `cron_schedule_file: "config/users_schedule.yml"`
+3. Load the file manually as follows
 ```ruby
 # config/initializers/sidekiq.rb
-schedule_file = "config/schedule.yml"
+schedule_file = "config/users_schedule.yml"
 
 if File.exist?(schedule_file) && Sidekiq.server?
   Sidekiq::Cron::Job.load_from_hash YAML.load_file(schedule_file)
```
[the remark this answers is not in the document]
sandip-mane: Updated it 👍

danilogco (approved):

sandip-mane:
@markets my pleasure🙌.  Thank you for reviewing it attentively.

Shall we release a new version of the gem? _(Because the README states that the file is loaded automatically.)_

markets:
@sandip-mane @microstudi We are preparing a new release here 👀 https://github.com/ondrejbartas/sidekiq-cron/pull/340

legendetm:
As we where using `erb` in the `schedule.yml` file, then this update broke the sidekiq setup. We had to rename the file as `config/schedule.yml.erb` to skip autoloading.

lib/sidekiq/cron/schedule_loader.rb
legendetm: You can skip automatic loader by renaming your `schedule.yml` file, or by setting `cron_schedule_file` path to non-existing file. And then load you schedule files manually.

qichunren:
I used the second way to integrate cron job, but does't work as expect. https://github.com/sidekiq-cron/sidekiq-cron?tab=readme-ov-file#loading-jobs-from-schedule-file, it says: multiple ways, so I used the second way to test.

This is my question here: https://github.com/sidekiq-cron/sidekiq-cron/discussions/560

PR #338, opened 2022-06-11
Fix Sidekiq.options deprecation warning

chrismaximin:
Since Sidekiq 6.5.0, using `Sidekiq.options[:config_var]` throws a warning: https://github.com/mperham/sidekiq/pull/5340/files#diff-2eecd3565b38823c09513903afa6686627d3ba8d883d520d3a83db61375e5ddeR72

> 2022-06-11T09:47:02.416Z pid=88697 tid=2k9 WARN: `config.options[:key] = value` is deprecated, use `config[:key] = value`: [".../gems/sidekiq-cron-1.5.1/lib/sidekiq/cron/job.rb:481:in `add_jid_history'", ".../gems/sidekiq-cron-1.5.1/lib/sidekiq/cron/job.rb:73:in `enque!'"]

This change uses the new API, if it's available.

Before (mperham/sidekiq@eddc11ba): `Sidekiq.respond_to?(:[]) # => false`
After (mperham/sidekiq@67daa7a4): `Sidekiq.respond_to?(:[]) # => true`

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ For example: `"*/30 * * * * *"` would schedule a job to run every 30 seconds.
 Note that if you plan to schedule jobs with second precision you may need to override the default schedule poll interval so it is lower than the interval of your jobs:
 
 ```ruby
-Sidekiq.options[:average_scheduled_poll_interval] = 10
+Sidekiq[:average_scheduled_poll_interval] = 10
 ```
 
 The default value at time of writing is 30 seconds. See [under the hood](#under-the-hood) for more details.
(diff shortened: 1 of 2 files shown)


danilogco (approved):

danilogco:
Maybe we can create an alias to the Sidekiq configs in the `sidekiq-cron` module to avoid these multiples `if`s around the code. At least the `if` will be there only.

markets:
The @danilogco's proposal makes sense IMO, we can centralize this "compatibility" check in one place only.

markets (approved):

laertispappas:
Hello, can we have a patch released version for this fix?

markets:
@laertispappas preparing new release: https://github.com/ondrejbartas/sidekiq-cron/pull/340

laertispappas:
Thank you @markets much appreciated

PR #343, opened 2022-07-04
unify code comments style, fix typos, remove useless comments + update example image

markets:
- unify code comments style, fix typos and remove useless comments
- update WEB UI example image to more recent version


honzasterba (approved):
next stop: Rubocop?

danilogco:
@markets I usually adopt the rubocop pattern from the rails project adding some extensions (rubocop-i18n, rubocop-packaging, rubocop-performance, rubocop-rspec, rubocop-thread_safety). I 100% agree with that and would suggest to use the rubocop just after the refacs to make it compatible to the Sidekiq 7.

PR #359, opened 2022-09-08
Fix poller affecting Sidekiq scheduled set poller

### LINKED ISSUE #358: Sidekiq cron affects Sidekiq scheduled set polling interval
This is a bug introduced in https://github.com/ondrejbartas/sidekiq-cron/pull/331/files (v1.5.0).

Previously, we were overriding the `poll_interval_average` method in the poller class. This means it only affects `Sidekiq::Cron::Poller` which is what we want.

With that PR, we are now setting the `poll_interval_average` Sidekiq option, which means Sidekiq would be using this interval instead of the scaled interval based on number of Sidekiq processes: https://github.com/mperham/sidekiq/blob/f3e0640a79fec2ea9a87bdfab12e7c45b4153778/lib/sidekiq/scheduled.rb#L176-L178

engwan:
Avoid setting global config so that we do not change the behavior of the Sidekiq scheduled set poller.

Closes #358

### DIFF
--- a/lib/sidekiq/cron/poller.rb
+++ b/lib/sidekiq/cron/poller.rb
@@ -5,20 +5,17 @@ require 'sidekiq/options'
 
 module Sidekiq
   module Cron
-    POLL_INTERVAL = Sidekiq::Options[:average_scheduled_poll_interval] || 30
+    POLL_INTERVAL = 30
 
     # The Poller checks Redis every N seconds for sheduled cron jobs.
     class Poller < Sidekiq::Scheduled::Poller
-      def initialize
-        Sidekiq.configure_server do
-          Sidekiq::Options[:poll_interval_average] = POLL_INTERVAL
-        end
-
+      def initialize(options = {})
         if Gem::Version.new(Sidekiq::VERSION) >= Gem::Version.new('6.5.0')
-          # Sidekiq Poller init requires a config argument.
-          super(Sidekiq)
-        else
           super
+        else
+          # Old version of Sidekiq does not accept a config argument.
+          @config = options
+          super()
         end
       end
 
@@ -44,6 +41,10 @@ module Sidekiq
         Sidekiq.logger.error "CRON JOB: #{ex.message}"
         Sidekiq.logger.error "CRON JOB: #{ex.backtrace.first}"
         handle_exception(ex) if respond_to?(:handle_exception)
+      end
+
+      def poll_interval_average
+        @config[:average_scheduled_poll_interval] || POLL_INTERVAL
       end
     end
   end
--- a/lib/sidekiq/cron/launcher.rb
+++ b/lib/sidekiq/cron/launcher.rb
@@ -13,7 +13,7 @@ module Sidekiq
 
       # Add cron poller and execute normal initialize of Sidekiq launcher.
       def initialize(options)
-        @cron_poller = Sidekiq::Cron::Poller.new
+        @cron_poller = Sidekiq::Cron::Poller.new(options)
         super(options)
       end
 


lib/sidekiq/cron/poller.rb
```diff
@@ -5,20 +5,17 @@
...
+      def initialize(options)
         if Gem::Version.new(Sidekiq::VERSION) >= Gem::Version.new('6.5.0')
-          # Sidekiq Poller init requires a config argument.
-          super(Sidekiq)
-        else
           super
+        else
+          # Old version of Sidekiq does not accept a config argument.
+          @config = Sidekiq.options
+          super()
```
engwan, line 18: Newer versions of Sidekiq pass in the options to the poller so that we don't have to read the global config. I think this is for the future where you can run multiple Sidekiq instances side-by-side.

So instead of making this behave like the old poller, we should follow this new strategy and just fall back to reading the global config on older versions.

lib/sidekiq/cron/poller.rb
```diff
@@ -45,6 +42,10 @@ def enqueue_job(job, time = Time.now.utc)
         Sidekiq.logger.error "CRON JOB: #{ex.backtrace.first}"
         handle_exception(ex) if respond_to?(:handle_exception)
       end
+
+      def poll_interval_average
+        @config[:average_scheduled_poll_interval] || POLL_INTERVAL
```
engwan, line 47: I do not agree with using `average_scheduled_poll_interval` here because this makes it intertwined with the Sidekiq setting.

Default value for this in Sidekiq is actually 5 seconds. So when we introduced this option, we actually dropped the default interval to 5 seconds and the docs are wrong.

I think a better approach would be to use a different setting like `cron_poll_interval`. I kept this as-is in this PR though since this is a different bug.

markets: I agree with this, we should not change the Sidekiq settings... this will be surprising for most of the users and can potentially introduce weird bugs. It's ok to read Sidekiq settings to take, for example, default settings... but overriding those settings seems a bad idea.

engwan: Created a separate issue for this: https://github.com/ondrejbartas/sidekiq-cron/issues/361

markets: btw, where this method is called?

engwan: This is called in the parent class: https://github.com/mperham/sidekiq/blob/v6.4.2/lib/sidekiq/scheduled.rb#L150-L157

markets:
LGTM ✅, but I'd like to hear @danilogco thoughts before merge.

lib/sidekiq/cron/poller.rb
```diff
@@ -5,20 +5,17 @@
...
+    POLL_INTERVAL = 30
 
     # The Poller checks Redis every N seconds for sheduled cron jobs.
     class Poller < Sidekiq::Scheduled::Poller
-      def initialize
-        Sidekiq.configure_server do
-          Sidekiq::Options[:poll_interval_average] = POLL_INTERVAL
-        end
-
+      def initialize(options)
```
danilogco, line 12: I would like to suggest to make this parameter optional to make it a little more clean.

```suggestion
      def initialize(options = {})
```

test/integration/performance_test.rb
```diff
@@ -24,7 +24,7 @@
       Sidekiq::Cron::Job.create(args.merge(name: "Test#{i}"))
     end
 
-    @poller = Sidekiq::Cron::Poller.new
+    @poller = Sidekiq::Cron::Poller.new({})
```
danilogco, line 27: And then, once its optional, we could just init the poller using the old way.

```suggestion
    @poller = Sidekiq::Cron::Poller.new
```

engwan:
@danilogco I've updated with your suggestion. Can you take another look please?

danilogco (approved):

markets:
Thank you @engwan @danilogco! Merging 🚀

PR #360, opened 2022-09-08
Add Truffleruby support to CI

gogainda:


markets:
Thanks @gogainda but not interested at this moment. 

This is a Sidekiq extension, and Sidekiq itself doesn't run CI against that Ruby implementation, so it makes no sense to support it here.

Issue #361, opened 2022-09-09
Default polling interval is 5 seconds and not 30 seconds as documented


engwan:
From https://github.com/ondrejbartas/sidekiq-cron/pull/359#discussion_r965969273:

> I do not agree with using average_scheduled_poll_interval here because this makes it intertwined with the Sidekiq setting.
> 
> Default value for this in Sidekiq is actually 5 seconds. So when we introduced this option, we actually dropped the default interval to 5 seconds and the docs are wrong.
> 
> I think a better approach would be to use a different setting like cron_poll_interval. I kept this as-is in this PR though since this is a different bug.

markets:
hey @engwan now that #359 has been merged, are you interested in working on this?

PR #362, opened 2022-09-16
Fix default polling interval

engwan:
Uses a different Sidekiq option because `average_scheduled_poll_interval` has a default value of 5 seconds.

I know this is a breaking change but so was the bug that changed it to 5 seconds from 30 seconds.

Closes https://github.com/ondrejbartas/sidekiq-cron/issues/361


engwan:
What do you think of this @markets @danilogco?

honzasterba (approved):
Looks good.

PR #368, opened 2022-09-23
Allow disabling of cron polling

engwan:
When running with very many Sidekiq processes, polling on each one of them can be expensive and unnecessary too.

This allows disabling polling by setting `cron_poll_interval` to zero.


engwan:
@markets what do you think of this change? At GitLab.com, we have lots of Sidekiq processes and lots of cron jobs so we end up with lots of HGETALL calls which leads to unnecessary load on Redis.

More details in https://gitlab.com/gitlab-com/gl-infra/scalability/-/issues/1932

markets:
I like the idea to be able to disable polling for specific processes! We'll need to add some tests though.

engwan:
Yup. I've just added the tests.

markets (approved):

Issue #300, opened 2021-03-06
NoMethodError: undefined method `jid=' for


vitobotta:
Hi!

Thanks for this gem, very useful. I have been using it for a while and the only issue that I have is that quite often jobs enqueued fail with the error in the subject, but only in development. It doesn't seem to happen in production.

I am currently using Sidekiq 6.1.2. Is sidekiq-cron fully compatible with this version?

Thanks!

iHiD:
I'm also getting this. I've tried upgrading to the `6.2.1` and still get the same. @vitobotta Did you solve this?

vitobotta:
@iHiD Nope, still happening...

christopheragnus:
anyone able to solve this? are you guys using Activejob or Sidekiq Workers?

looks like there's a similar problem: https://github.com/mperham/sidekiq/issues/3593

iHiD:
No. I switched to `sidekiq-scheduler` instead. I couldn't make it stop with `sidekiq-cron`.

phlegx:
See https://github.com/mperham/sidekiq/wiki/Ent-Periodic-Jobs#notes
Cron jobs must use `Sidekiq::Job`, not `ActiveJob`.

PR #369, opened 2022-09-29
Support Sidekiq 7

serprex:
Sidekiq v7 support 🎉 

Closes #378
Closes #370
Closes #357

PR #381, opened 2022-11-29
always enque via active_job interface when defined in cron job config

hqm42:
This should help to prevent #300 by setting `active_job: true` in config/schedule.yml


andreavocado (commented):
Looks good to me 👍

markets:
Seems fine thanks! I'll be back to this one after Sidekiq v7 support (#369) is merged (we're so close!).

hqm42:
Another aproch for preventing #300 would be to ducktype the enque method:
```ruby
if klass_const
  if klass_const.respond_to?(:perform_later)
    enqueue_active_job # ...
  elsif klass_const.respont_to?(:perform_async)
    enqueue_sidekiq_worker # ...
  else
    # log error / raise
  end
end
```

markets:
@hqm42 could you please update from `master`?

markets (approved):

Issue #387, opened 2022-12-13
before_worker_boot failed with exception (ArgumentError)  using Sidekiq v7 on embedded mode


eddygarcas:
Ruby version: 3.1.2
Rails version: "~> 7.0.3"
Sidekiq / Pro / Enterprise version(s): 7.0.1
Sidekiq-cron: master

Hi, hope you can advice me on this.
I'm currently using the latest Sidekiq 7 on embedded mode as i don't what to run multiple sidekiq instances on the same app.

Following the steps described [here](https://github.com/mperham/sidekiq/wiki/Embedding#puma) to enabled embedded mode on a Puma server and running docker I've got this trace:
```
12:27:45 web.1  | [8353] ! WARNING: Detected running cluster mode with 1 worker.
12:27:45 web.1  | [8353] ! Running Puma in cluster mode with a single worker is often a misconfiguration.
12:27:45 web.1  | [8353] ! Consider running Puma in single-mode (workers = 0) in order to reduce memory overhead.
12:27:45 web.1  | [8353] ! Set the `silence_single_worker_warning` option to silence this warning message.
12:27:45 web.1  | DEBUGGER: Attaching after process 8353 fork to child process 8364
12:27:45 web.1  | Subprocess Debugger (ruby-debug-ide 0.7.3, debase 0.2.5.beta2, file filtering is supported) listens on 0.0.0.0:36159
12:28:45 web.1  | [8353] ! Terminating timed out worker (worker failed to boot within 60 seconds): 8364
12:28:45 web.1  | DEBUGGER[bin/rails#8441]: Attaching after process 8353 fork to child process 8441
12:28:45 web.1  | Subprocess Debugger (ruby-debug-ide 0.7.3, debase 0.2.5.beta2, file filtering is supported) listens on 0.0.0.0:43215
12:29:14 web.1  | [8441] WARNING hook before_worker_boot failed with exception (ArgumentError) wrong number of arguments (given 2, expected 1)
```
I've added an issue in Sidekiq but seems that the issue relies on Sidekiq-cron, as per following trace:

```
/3.1.0/bundler/gems/sidekiq-cron-70ce5c57ab36/lib/sidekiq/cron/launcher.rb:17:in 'initialize'
/3.1.0/gems/sidekiq-7.0.1/lib/sidekiq/embedded.rb:16:in 'new'
[...]

PR #389, opened 2022-12-22
Fix ArgumentError with embedded mode, sidekiq 7.0 instantiating launcher with unexpected signature

rromanchuk:
PR for reference #387 

Feel free to close, haven't followed up and may already be fixed.

### DIFF
--- a/lib/sidekiq/cron/launcher.rb
+++ b/lib/sidekiq/cron/launcher.rb
@@ -14,7 +14,7 @@ module Sidekiq
       attr_reader :cron_poller
 
       # Add cron poller and execute normal initialize of Sidekiq launcher.
-      def initialize(config, embeded: true)
+      def initialize(config, embedded: true)
         config[:cron_poll_interval] = DEFAULT_POLL_INTERVAL if config[:cron_poll_interval].nil?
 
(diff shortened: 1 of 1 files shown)


markets:
It seems people make it work with a workaround: https://github.com/sidekiq-cron/sidekiq-cron/issues/387#issuecomment-1352787715

Maybe this change will make it incompatible with previous Sidekiq versions?

mikebaldry:
The original sidekiq signature is `config, embedded: false` - defaulting it to `embedded: true` is probably not desirable and if running in embedded mode, it would be `true` anyway. Maybe to keep it backwards compatible, `config, *` would be better.

```ruby

class OldLauncher
  def initialize(config)
    puts "old: #{config}"
  end
end

class NewLauncher
  def initialize(config, embedded: false)
    puts "new: #{config} embedded: #{embedded}"
  end
end

module CronPrepend
  def initialize(config, *)
    puts "cron prepend"
    super
  end
end

OldLauncher.prepend(CronPrepend)
NewLauncher.prepend(CronPrepend)

> OldLauncher.new(123)
cron prepend
old: 123
> NewLauncher.new(123)
cron prepend
new: 123 embedded: false
> NewLauncher.new(123, embedded: true)
cron prepend
new: 123 embedded: true
```

PR #392, opened 2023-01-31
fix argument error problem update from 1.6.0 to newer

mkusaka:
Fixed an error related to `date_as_argument` option that occurred when temporarily deploying multiple versions of sidekiq-cron.

This problem affects when you have a deployment where multiple versions of sidekiq-cron temporarily exist (such as [rolling update](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/deployment-type-ecs.html))

In this case, an option created in a newer version was unintentionally processed in the older version (1.6.0) and did not work correctly.

For more information, I have attempted to explain this at https://github.com/sidekiq-cron/sidekiq-cron/issues/350#issuecomment-1409798837

This PR avoids this problem by changing the hash of options created in the new version to work in the 1.6.0 version as well.

close: https://github.com/sidekiq-cron/sidekiq-cron/issues/350

### DIFF
--- a/lib/sidekiq/cron/job.rb
+++ b/lib/sidekiq/cron/job.rb
@@ -398,13 +398,12 @@ module Sidekiq
 
       # Export job data to hash.
       def to_hash
-        {
+        h = {
           name: @name,
           klass: @klass.to_s,
           cron: @cron,
           description: @description,
           args: @args.is_a?(String) ? @args : Sidekiq.dump_json(@args || []),
-          date_as_argument: date_as_argument? ? "1" : "0",
           message: @message.is_a?(String) ? @message : Sidekiq.dump_json(@message || {}),
           status: @status,
           active_job: @active_job ? "1" : "0",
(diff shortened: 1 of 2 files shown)


markets:
@vsuhachev could you please review or leave your thoughts this one 🙏🏼?

lib/sidekiq/cron/job.rb
```diff
@@ -398,13 +398,12 @@ def jid_history_from_redis
 
       # Export job data to hash.
       def to_hash
-        {
+        h = {
```
markets, line 401: maybe `hash` is more descriptive?

lib/sidekiq/cron/job.rb
```diff
@@ -413,6 +412,12 @@ def to_hash
           last_enqueue_time: @last_enqueue_time.to_s,
           symbolize_args: symbolize_args? ? "1" : "0",
         }
+
+        if date_as_argument?
```
markets, line 416: could we add a test to cover this behaviour?

lib/sidekiq/cron/job.rb
mkusaka: Thanks! fixed at ae56644

lib/sidekiq/cron/job.rb
mkusaka: I didn't know the best way to do this, but I added it! 503473b

markets (approved):

PR #396, opened 2023-02-12
Destroy all scheduled jobs before load a new on startup

dapi:
The problem: Sidekiq::Cron continue to store tasks that are no longer in the default config (config/schedule.yml
)
Solution: Destry all old jobs from cron before load new.

### DIFF
--- a/lib/sidekiq/cron/schedule_loader.rb
+++ b/lib/sidekiq/cron/schedule_loader.rb
@@ -10,8 +10,10 @@ if Sidekiq.server?
       config.on(:startup) do
         schedule = Sidekiq::Cron::Support.load_yaml(ERB.new(IO.read(schedule_file)).result)
         if schedule.kind_of?(Hash)
+          Sidekiq::Cron::Job.destroy_all!
           Sidekiq::Cron::Job.load_from_hash schedule
         elsif schedule.kind_of?(Array)
+          Sidekiq::Cron::Job.destroy_all!
           Sidekiq::Cron::Job.load_from_array schedule
         else
           raise "Not supported schedule format. Confirm your #{schedule_file}"


markets:
Hello @dapi 👋🏼

That makes sense to me! Let me ask @sandip-mane and @machisuke as they were both implied in the implementation of this feature.

sandip-mane:
@markets I had this doubt in the main PR itself. Ref: https://github.com/sidekiq-cron/sidekiq-cron/pull/337#issuecomment-1160347813

I guess using just the bang versions of these methods should fix the issue.

https://github.com/sidekiq-cron/sidekiq-cron/blob/71c02bfbe91ee1c22acb768b5fc0de4c280357dc/lib/sidekiq/cron/schedule_loader.rb#L12-L18

markets:
> I guess using just the bang versions of these methods should fix the issue.

Oh yes! You're right @sandip-mane! Thanks for your review!

markets:
@dapi could you please update the patch? The bang! versions should do the job.

markets:
hey @dapi 👋🏼 Did you have the chance to review those comments? I'd like to include this patch in the next release.

sandip-mane:
I can update the PR if it's urgent. When is the release planned for?

markets:
Thanks @sandip-mane! We still don't have a date planned for the release, but we started to work on it.

sandip-mane:
@markets I've raised a new PR #405 as I was not able to update this.

markets:
Fixed in https://github.com/sidekiq-cron/sidekiq-cron/commit/48c2563a77fb9dac4ef4e881bf934ae67f1f42ab

zaratan:
Hey ! I know this is a old issue but would that be ok for you if it was possible to remove this with a configuration flag or env variable ? We manage our cron-scheduled jobs both in the schedule.yml and through code and every reboot/redeploy our code managed cron-scheduled jobs are removed. I pinned the version of sidekiq-cron to `1.9.1` because of this.

zaratan:
I don't mind making the PR for this.

markets:
Hello @zaratan 👋🏼  

If you want to avoid the auto-reload of the schedule, you can just use a different YAML file name:

https://github.com/sidekiq-cron/sidekiq-cron/blob/aacd84bd08032b63917bb354e8b9607386388807/lib/sidekiq/cron/schedule_loader.rb#L9

This will avoid to enter that `if` code path and call the `Sidekiq::Cron::Job.load_from_hash!`. 

See more in the original PR: https://github.com/sidekiq-cron/sidekiq-cron/pull/337

zaratan:
Using a different YAML filename and loading manualy with `load_from_hash` should work indeed. Thanks !

PR #398, opened 2023-02-27
Allow for keyword args in Poller

samsm:
Formerly, `Sidekiq::Launcher.new` took one "config" argument. As of v7 there's a possibility of  "embedded" being passed in. ex: `Sidekiq::Launcher.new(config, embedded: true)`

Presently, when running sidekiq-cron with embedded mode, we get an exception. To reproduce this, load sidekiq-cron with Sidekiq in "[embedded mode](https://www.mikeperham.com/2022/10/27/sidekiq-7.0-embedding/)"

This PR adds optional keyword args to pass along, to support `embedded` and to allow for future keyword args that might get added.

### DIFF
--- a/lib/sidekiq/cron/launcher.rb
+++ b/lib/sidekiq/cron/launcher.rb
@@ -14,7 +14,7 @@ module Sidekiq
       attr_reader :cron_poller
 
       # Add cron poller and execute normal initialize of Sidekiq launcher.
-      def initialize(config)
+      def initialize(config, **kwargs)
         config[:cron_poll_interval] = DEFAULT_POLL_INTERVAL if config[:cron_poll_interval].nil?
 
         @cron_poller = Sidekiq::Cron::Poller.new(config) if config[:cron_poll_interval] > 0


markets:
Hi @samsm

We'll need to be sure this change is compatible Sidekiq <7.

samsm:
Hmm, do you mean it crashes on Ruby 2.6? Sidekiq 6 seems to handle this change ok. 

I know you can have whatever support policy you want for sidekiq-cron, but will note that Ruby 2.6 is past end of life and no longer officially supported: https://www.ruby-lang.org/en/downloads/branches/

samsm:
@markets Given it's an old Ruby version, perhaps this condition on those tests is an ok compromise?

samsm:
@markets I'll offer that this test locally passed on Ruby 3 with Sidekiq 6, but I agree there's no easy way to enshrine that in the test suite without the CI change you're mentioning 😅 

The current version in this PR skips. those tests on Ruby 2.6.

markets:
ok @samsm thanks for testing! 

I'm ok to merge, but I'd feel more confortable having Sidekiq v6 testing in our CI matrix (specially to cut a new release). There is a similar setup here: 

https://github.com/sidekiq-scheduler/sidekiq-scheduler/blob/f772e4f702ec7e545cf9cf7a7692f9b7b9a21d3d/.github/workflows/ci.yml#L15 

We can use it 👀 for inspiration.

markets:
@samsm Started to work on that in #399. Hopefully will be merged soon and we'll be able to automatically check different Sidekiq versions on CI.

markets:
@samsm https://github.com/sidekiq-cron/sidekiq-cron/pull/399 has been merged, could you please udpate/rebase?

samsm:
@markets Just did!

test/unit/launcher_test.rb
```diff
@@ -4,6 +4,7 @@
   describe 'initialization' do
     before do
       Sidekiq::Options[:cron_poll_interval] = nil
+      skip("This test does not work on Ruby 2.6.X") if RUBY_VERSION.index('2.6.') == 0
```
markets, line 7: ✂️

markets: @samsm We can remove this line now

samsm: @markets Done

markets (approved):
Thanks @samsm!

PR #405, opened 2023-03-03
Clear old jobs while loading the jobs from schedule

sandip-mane:
Patch for #396

### DIFF
--- a/test/unit/job_test.rb
+++ b/test/unit/job_test.rb
@@ -1273,11 +1273,11 @@ describe "Cron Job" do
       end
 
       it "duplicate jobs are not loaded" do
-        out = Sidekiq::Cron::Job.load_from_hash @jobs_hash
+        out = Sidekiq::Cron::Job.load_from_hash! @jobs_hash
         assert_equal out.size, 0, "should have no errors"
         assert_equal Sidekiq::Cron::Job.all.size, 2, "Should have 2 jobs after load"
 
-        out_2 = Sidekiq::Cron::Job.load_from_hash @jobs_hash
+        out_2 = Sidekiq::Cron::Job.load_from_hash! @jobs_hash
         assert_equal out_2.size, 0, "should have no errors"
         assert_equal Sidekiq::Cron::Job.all.size, 2, "Should have 2 jobs after loading again"
       end
--- a/lib/sidekiq/cron/schedule_loader.rb
+++ b/lib/sidekiq/cron/schedule_loader.rb
@@ -10,9 +10,9 @@ if Sidekiq.server?
       config.on(:startup) do
         schedule = Sidekiq::Cron::Support.load_yaml(ERB.new(IO.read(schedule_file)).result)
         if schedule.kind_of?(Hash)
-          Sidekiq::Cron::Job.load_from_hash schedule
+          Sidekiq::Cron::Job.load_from_hash! schedule
         elsif schedule.kind_of?(Array)
-          Sidekiq::Cron::Job.load_from_array schedule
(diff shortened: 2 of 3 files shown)


sandip-mane:
@markets I was not able to quickly update PR #396, hence created a new one. Please review.

markets:
Thanks @sandip-mane! Maybe we should update some tests too? I think the related tests are still using the non-bang version.

test/unit/job_test.rb
```diff
@@ -1273,11 +1273,11 @@
       end
 
       it "duplicate jobs are not loaded" do
-        out = Sidekiq::Cron::Job.load_from_hash @jobs_hash
+        out = Sidekiq::Cron::Job.load_from_hash! @jobs_hash
         assert_equal out.size, 0, "should have no errors"
         assert_equal Sidekiq::Cron::Job.all.size, 2, "Should have 2 jobs after load"
 
-        out_2 = Sidekiq::Cron::Job.load_from_hash @jobs_hash
+        out_2 = Sidekiq::Cron::Job.load_from_hash! @jobs_hash
```
markets, line 1280: I was referring the `ScheduleLoader` tests: https://github.com/sidekiq-cron/sidekiq-cron/blob/master/test/unit/schedule_loader_test.rb

sandip-mane: Yes, seemed like the right thing to do. Have updated the tests.

markets (approved):

fabn:
This should have been a breaking change, I used to create some cronjobs programmatically and they are gone on every restart.

rchasman (commented):
This PR introduces an issue for us, where it destroys existing jobs when we deploy. We don't load our schedules from a yml, they are dynamically created and stored in redis.

sandip-mane:
@markets shall we revert this change? The issue with dynamic cron jobs will continue to be an issue on every deploy IMO.
But it also leads to another issue where if I modify the class for an existing cron job, then I have to manually delete the old one using a script.

I think the best way would be to document the usage of dynamic cron jobs.

Thoughts?

sandip-mane:
> We only load some of our schedules from a YAML, because of this, the ones that are dynamically created and stored in Redis are cleared.
 
@rchasman are all jobs from `ScheduledSet` cleared or only the Cron jobs are cleared?

markets:
hmm 🤔 maybe we should revert this change... It seems a good idea, but it seems it breaks with dynamic jobs, which is worst scenario compared to old jobs in the schedule (that can be cleaned manually).

Another idea (probably ideal solution?) is to only cleanup the ones defined in the YAML, but right now, we don't have a mechanism (an attribute or an extra data structure) to distinct between dynamic jobs vs static (defined in the YAML). 

So yes, maybe it's better to revert it and document it?

sandip-mane:
**My honest opinion:** 
Cron schedule should be kept up to date upon every deployment. (For various reasons, for example, I rename an existing cronjob class, delete a cron-job etc)
This creates an issue with dynamic cron jobs, so they need to be handled separately. I have not worked on these yet so not sure what could be the best way to deal with them.
Probably having an attribute for dynamic jobs could help distinguish between the jobs and we will be able to handle it from the gem.

**I feel this should be the plan:**
1. Revert & release
2. Have a way to distinguish dynamic crons
3. Add this change back (it should only clear non-dynamic crons)

markets:
Agree with the plan! Please, feel free to send PRs to move this forward, unfortunately I've very limited time at this point to work on these changes.

sandip-mane:
Sure sure, I will try to send PRs for the other 2 issues by this weekend (Working on a release towards a deadline).

PR #406, opened 2023-03-03
Remove ✂️ Docker support for development

markets:
I started a discussion some months ago here with no success 😅: https://github.com/sidekiq-cron/sidekiq-cron/discussions/383

I'm not sure if they actually work (it uses an EOL Ruby `FROM ruby:2.6`) or even how to run tests with them (there is no documentation about it). So maybe we can just ✂️ these files?


honzasterba (approved):

zedtux:
Too bad, I missed that one! I am using it for my devs, my machine is almost containers only (only the system is on an A/B partitioning).

I can update the `Dockerfile` in order to use an up-to-date Ruby version.

PR #409, opened 2023-04-26
Add the ability to deactivate the default schedule loader

dalpo:
Hello and thank you for this awesome library!

I drafted this PR because I'm having some troubles with my current sidekiq cron configuration and I hope it would be useful to someone else.

In the application I'm working in, I have many yaml schedule files (divided by application domain). 
But the current [schedule_loader](https://github.com/sidekiq-cron/sidekiq-cron/blob/master/lib/sidekiq/cron/schedule_loader.rb#L13) is deleting all previously configured jobs, keeping the only configurations inside the default `config/schedule.yml`file. 

NOTE: I simply load multiple yaml files inside the sidekiq initializer, as reported by the README:

Screenshot 2023-04-26 at 11 40 48

---

In my option the easiest way to handle this use-case, would be to [remove the automatic initialization of the schedule_loader](https://github.com/sidekiq-cron/sidekiq-cron/blob/master/lib/sidekiq/cron.rb#L4) and allow the user to manually require it inside its own sidekiq initializer:

```
# config/initializers/sidekiq.rb
require "sidekiq"
require "sidekiq/cron"
require "sidekiq/cron/schedule_loader" # optional
```

But this would mean to introduce a potential breaking change in the next release..

So, I drafted this PR to introduce a new configuration `enable_default_cron_scheduler` to enable/disable the schedule_loader initialization, by simply configure the application sidekiq.yml file.

Le me know you think, so that I could possibly improve this PR or just close it 😄

### DIFF
--- a/lib/sidekiq/cron.rb
+++ b/lib/sidekiq/cron.rb
@@ -5,5 +5,8 @@ require "sidekiq/cron/schedule_loader"
 
 module Sidekiq
   module Cron
+    def self.enabled_schedule_loader?
+      Sidekiq::Options.fetch(:enable_default_cron_schedule, true) == true
+    end
   end
 end
--- a/lib/sidekiq/cron/schedule_loader.rb
+++ b/lib/sidekiq/cron/schedule_loader.rb
@@ -2,13 +2,14 @@ require 'sidekiq'
 require 'sidekiq/cron/job'
 require 'sidekiq/options'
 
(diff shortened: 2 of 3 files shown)


dalpo:
I self close the PR. 

The different schedule files are now loading correctly.
I had trivially forgotten to specify the `config.on(:startup) do` block inside the sidekiq initializer 😄

PR #410, opened 2023-05-02
Use hset instead of deprecated hmset

haffla:
Hi.

Sidekiq now prints a warning about the usage of `hmset` when using sidekiq-cron. This Redis command [was deprecated in Redis v4](https://redis.io/commands/hmset/).

sidekiq-cron requires Sidekiq >= 6 which itself requires Redis >= 4. So I would argue that we can safely use `hset` (and go up a minor version).

I also don't see the need to convert the hash to an array. Unless there was a specific reason for that?

### DIFF
--- a/lib/sidekiq/cron/job.rb
+++ b/lib/sidekiq/cron/job.rb
@@ -469,7 +469,7 @@ module Sidekiq
           conn.sadd self.class.jobs_key, [redis_key]
 
           # Add informations for this job!
-          conn.hmset redis_key, *hash_to_redis(to_hash)
+          conn.hset redis_key, to_hash.transform_values! { |v| v || "" }
 
           # Add information about last time! - don't enque right after scheduler poller starts!
           time = Time.now.utc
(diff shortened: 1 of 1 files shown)


markets:
Thanks for this one @haffla! I'm ok with the change.

> I also don't see the need to convert the hash to an array. Unless there was a specific reason for that?

Can't see the need either... navigating through the git history, it seems it was introduced around 10 years ago, in the initial versions of the gem: https://github.com/sidekiq-cron/sidekiq-cron/commit/10e64d43abac58380b85c3b8959771994618853d. So probably we are fine ✂️.

@honzasterba @ondrejbartas any thoughts?

honzasterba:
looks good to me

honzasterba (approved):
👍🏻

spickermann:
Are there any plans to release a version `1.10.1` with this fix?

markets:
@spickermann I'll prepare a release for next week

PR #414, opened 2023-05-24
Reduce gem size by excluding test files

ybiquitous:
```console
$ gem build -o before

$ git switch reduce-gem-size

$ git build -o after

$ du -sh before after
 32K	before
 24K	after
```

=> **-8K** saved


markets (approved):

ybiquitous:
Thanks for the quick merge! 😄

PR #431, opened 2023-10-27
Differentiates b/w schedule vs dynamic jobs + Clears scheduled jobs upon schedule load

### LINKED ISSUE #428: Add ability to differentiate between regular cron jobs and dynamic cron jobs
Ref: https://github.com/sidekiq-cron/sidekiq-cron/pull/405#issuecomment-1772568927

Suggestion: Add an attribute in regular cron jobs when they are loaded from the schedule file, using this, we will be able to differentiate between dynamically created cron jobs.

### LINKED ISSUE #429: Clear non-dynamic cron jobs upon loading the schedule
Ref: https://github.com/sidekiq-cron/sidekiq-cron/pull/405#issuecomment-1772568927

Work on this can be started when #428 is fixed.

sandip-mane:
fixes #428 
fixes #429

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ gem "sidekiq-cron"
 
 ## Getting Started
 
- ### Job properties
+### Job properties
 
 ```ruby
 {
@@ -51,6 +51,7 @@ gem "sidekiq-cron"
   'cron' => '1 * * * *',  # execute at 1 minute of every hour, ex: 12:01, 13:01, 14:01, ...
   'class' => 'MyClass',
   # OPTIONAL
+  'source' => 'dynamic', # source of the job, `schedule`/`dynamic` (default: `dynamic`)
   'queue' => 'name of queue',
   'args' => '[Array or Hash] of arguments which will be passed to perform method',
   'date_as_argument' => true, # add the time of execution as last argument of the perform method
@@ -223,7 +224,10 @@ array = [
 Sidekiq::Cron::Job.load_from_array array
 ```
 
-Bang-suffixed methods will remove jobs that are not present in the given hash/array, update jobs that have the same names, and create new ones when the names are previously unknown.
+Bang-suffixed methods will
+remove jobs where source is `schedule` and are not present in the given hash/array,
+update jobs that have the same names,
+and create new ones when the names are previously unknown.
 
 ```ruby
 Sidekiq::Cron::Job.load_from_hash! hash
@@ -262,7 +266,9 @@ Sidekiq.configure_server do |config|
     schedule_file = "config/users_schedule.yml"
 
     if File.exist?(schedule_file)
-      Sidekiq::Cron::Job.load_from_hash YAML.load_file(schedule_file)
+      schedule = YAML.load_file(schedule_file)
+
+      Sidekiq::Cron::Job.load_from_hash!(schedule, source: "schedule")
     end
   end
 end
@@ -314,6 +320,17 @@ job.status
 # enqueue job right now!
 job.enque!
 ```
+
+### Schedule vs Dynamic jobs
+
+There are 2 possible sources for jobs i.e. `schedule/dynamic`.
+Jobs that are loaded with schedule files have source as `schedule`,
+while the jobs that are created at runtime without `source=schedule` argument will have source as dynamic.
+
+The major difference is that, upon loading the schedule,
+stale `schedule` jobs are destroyed,
+this will keep the jobs in the schedule in sync.
+The `dynamic` jobs are not affected by this.
 
 ### How to start scheduling?
 
--- a/README.md
+++ b/README.md
@@ -136,6 +136,8 @@ which will ensure that arguments you are passing to it will be symbolized when p
 
 #### Adding Cron job
 
+Refer to [Schedule vs Dynamic jobs](#schedule-vs-dynamic-jobs) to understand the difference.
+
 ```ruby
 class HardWorker
   include Sidekiq::Worker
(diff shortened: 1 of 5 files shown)


sandip-mane:
@markets please review.

lib/sidekiq/cron/job.rb
```diff
@@ -197,7 +197,7 @@ def self.load_from_hash! hash
       def self.load_from_array array
         errors = {}
         array.each do |job_data|
-          job = new(job_data)
+          job = new(job_data.merge(source: "schedule"))
```
sandip-mane, line 200: I had 2 options, `regular` and `schedule`. I chose `schedule` because that goes well with `source`.

markets:
Thanks for working on this @sandip-mane!

I think we should change a bit how with set the `source: "schedule"`. Right now, you added it in the `load_from_array` method, but this can cause problems:
- People calling this method from code will set "schedule" for dynamic jobs
- People using the `schedule_loader` with a hash (https://github.com/sidekiq-cron/sidekiq-cron/blob/master/lib/sidekiq/cron/schedule_loader.rb#L13)

So, not sure if I'm missing something, but I think the correct way should be something like: merge the `source: "schedule"` pair only in the `schedule_loader` code, like: `Sidekiq::Cron::Job.load_from_hash schedule, { source: "schedule" }`. Everything else should be marked dynamic right?

sandip-mane:
Yes, makes sense. I'll make the change.

Sent from Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

sandip-mane:
> - People calling this method from code will set "schedule" for dynamic jobs

I had considered this use case, and assumed that folks would be using `Sidekiq::Cron::Job.create` or `Sidekiq::Cron::Job.new().save`. I might be wrong because I have not used dynamic jobs yet.

> - People using the schedule_loader with a hash (https://github.com/sidekiq-cron/sidekiq-cron/blob/master/lib/sidekiq/cron/schedule_loader.rb#L13)

`load_from_hash` method internally uses `load_from_array` itself. Hence, `load_from_array` seems single source of truth. Ref: [load_from_hash](https://github.com/sandip-mane/sidekiq-cron/blob/c020a4a4562cfe7ffe21ffdfb8175bee5a87915f/lib/sidekiq/cron/job.rb#L170)

### One problem with the new approach:

> Sidekiq::Cron::Job.load_from_hash schedule, { source: "schedule" }

There's still one problem with this approach. 
Some people might be using their custom schedule loading mechanism, in that case, their actual scheduled jobs will still have `source=dynamic`. 

Thoughts?

sandip-mane:
@markets missed mentioning you above.

sandip-mane:
@markets 

I like the idea of setting the default to `schedule`. But, that way, most people, will have to make some changes to use the `dynamic` cronjobs and until they do, they will keep on losing the jobs. (As we are clearing the scheduled jobs on load) Assume someone using dynamic jobs in an old implementation upgrades the gem to the latest version, they will lose all their dynamic jobs, without any deprecation warning etc. 

I have used `dynamic` as default so that when people migrate, their existing cronjobs are not lost.

---

#### I have another approach in mind:

Currently, we have `load_from_hash!` and `load_from_array!` methods, these are being used for loading the schedule from the loader script here:

https://github.com/sidekiq-cron/sidekiq-cron/blob/c020a4a4562cfe7ffe21ffdfb8175bee5a87915f/lib/sidekiq/cron/schedule_loader.rb#L12-L18

We can create 2 new methods, i.e. `load_schedule_from_hash!` and `load_schedule_from_array!` which will default to loading the jobs as `schedules`. We can then add these to README for loading schedules from files.

This way, even if people don't use new methods that add jobs as `schedule`, they will not lose their jobs that are loaded with `load_from_array!/load_from_hash!`.


Please suggest.

markets:
Ok! I 100% agree with this 👇🏼 That's a really important point and we don't want to broke this again.

> Assume someone using dynamic jobs in an old implementation upgrades the gem to the latest version, they will lose all their dynamic jobs, without any deprecation warning etc.

About the approach with new methods: I think we are not fixing so much (they are quite similar to call it with the param `{ source: "schedule" }`) and we're introducing new public API.

Probably going with "dynamic" by default seems safe approach. So I think we can keep that, but instead of using a default value in the method, let's pass it from the scheduler loader `{ source: "schedule" }`. Additionally, let's add docs in the README about this new attribute (https://github.com/sidekiq-cron/sidekiq-cron?tab=readme-ov-file#job-properties), the default value, what it does, ... Around here seems also a good place to explain it for people going with option 3:

Captura de pantalla 2023-10-30 a las 19 50 01

What do you think?

sandip-mane:
Sounds good. I will make these changes. 👍

sandip-mane:
@markets I've added the suggested changes, please review.

README.md
```diff
@@ -223,7 +226,10 @@ array = [
 Sidekiq::Cron::Job.load_from_array array
 ```
 
-Bang-suffixed methods will remove jobs that are not present in the given hash/array, update jobs that have the same names, and create new ones when the names are previously unknown.
+Bang-suffixed methods will
+remove jobs where source is `schedule` and are not present in the given hash/array,
+update jobs that have the same names,
+and create new ones when the names are previously unknown.
```
markets, line 232: Why this break lines here?

lib/sidekiq/cron/job.rb
```diff
@@ -194,21 +194,21 @@ def self.load_from_hash! hash
       #   }
       # ]
       #
-      def self.load_from_array array
+      def self.load_from_array(array, options = {})
         errors = {}
         array.each do |job_data|
-          job = new(job_data)
+          job = new(job_data.merge(options.with_indifferent_access))
```
markets, line 200: Do we need `with_indifferent_access` here?

markets:
@sandip-mane just added a couple of minor comments, but overall looking good 👍🏼

README.md
sandip-mane: I am used to the [ventilated prose](https://vanemden.wordpress.com/2009/01/01/ventilated-prose/) writing style, I can revert this if need be.

lib/sidekiq/cron/job.rb
sandip-mane: Just noticed that we are transforming the keys into strings.

https://github.com/sandip-mane/sidekiq-cron/blob/f9335c89c92fa4f7872e2789de978fb8c1988e3d/lib/sidekiq/cron/job.rb#L273

I will revert this.

README.md
markets: Yes, nothing against this style, but it will be more consistent with the rest of the docs.

sandip-mane: Have removed this styling.

sandip-mane:
@markets I have made the changes.

markets:
Thanks @sandip-mane! I'm going to prepare a new release 🚀

jalada:
Hey gang, this really caught us out because it was added in a minor version bump, but the fact that the newly optional `source: 'dynamic'` or `source: 'schedule'` affects whether bang-suffixed methods delete old jobs is actually a breaking change - our schedule-loading code stopped deleting old jobs.

I think this should've been a `2.0.0` instead of a `1.11.0`.

PR #436, opened 2023-11-20
remove Sidekiq.server? check from schedule loader

oozzal:
Starting with 7.0, Sidekiq can be run in embedded mode where `Sidekiq.server?` returns `false`

Reference: https://www.mikeperham.com/2022/10/27/sidekiq-7.0-embedding/

### DIFF
--- a/lib/sidekiq/cron/schedule_loader.rb
+++ b/lib/sidekiq/cron/schedule_loader.rb
@@ -2,20 +2,18 @@ require 'sidekiq'
 require 'sidekiq/cron/job'
 require 'sidekiq/options'
 
-if Sidekiq.server?
-  Sidekiq.configure_server do |config|
-    schedule_file = Sidekiq::Options[:cron_schedule_file] || 'config/schedule.yml'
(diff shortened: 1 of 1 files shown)


markets:
hey @oozzal 👋🏼 

Do you think this will work well or cause any problem under Sidekiq 6? We still support both versions:

https://github.com/sidekiq-cron/sidekiq-cron/blob/78eb493adbfeb908861ec9b3a76e5e7a01adf339/.github/workflows/ci.yml#L12

markets:
@sandip-mane are you agree with this changes?

sandip-mane:
@markets YES, should not cause any issues as we use Sidekiq's `config.on(:startup)` callback hence it is run only when Sidekiq boots.

I have just verified this change with `Sidekiq 6.5` and `7.1` and it worked as expected. Also verified it in `embedded` mode, and it works fine that way as well(similar to running multiple sidekiq processes). 

This is a non-critical change though, only applies when people run sidekiq in embedded mode(rare)

---

@oozzal thank you for the addition.

PR #438, opened 2023-11-27
Only check out a Redis connection if necessary

stanhu:
In Sidekiq 7, calls to `Sidekiq.redis` checks out a connection from an internal connection pool instead of the connection pool used by the job processors. If a connection is checked out but not used, it's possible that the heartbeat thread won't keep the connection alive before the Redis server client timeout.

To avoid this, update the checks to return earlier if the Redis connection is not needed.

This might help reduce intermittent `Broken pipe` errors as reported in https://github.com/redis-rb/redis-client/issues/119.

### DIFF
--- a/lib/sidekiq/cron/job.rb
+++ b/lib/sidekiq/cron/job.rb
@@ -21,11 +21,12 @@ module Sidekiq
 
       # Crucial part of whole enqueuing job.
       def should_enque? time
+        return false unless status == "enabled"
+        return false unless not_past_scheduled_time?(time)
+        return false unless not_enqueued_after?(time)
+
         enqueue = Sidekiq.redis do |conn|
-          status == "enabled" &&
-            not_past_scheduled_time?(time) &&
-            not_enqueued_after?(time) &&
-            conn.zadd(job_enqueued_key, formatted_enqueue_time(time), formatted_last_time(time))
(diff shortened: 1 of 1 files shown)


lib/sidekiq/cron/job.rb
```diff
@@ -239,12 +241,11 @@ def self.count
       def self.find name
         # If name is hash try to get name from it.
         name = name[:name] || name['name'] if name.is_a?(Hash)
+        return unless exists? name
 
         output = nil
         Sidekiq.redis do |conn|
-          if exists? name
```
stanhu, line 245: I wonder if this could cause a deadlock since `exists?` also checks out a connection.

markets (approved):
Thanks @stanhu!

lib/sidekiq/cron/job.rb
stanhu: Probably not, since `connection_pool` appears to reuse the same connection for a given thread.

stanhu:
@markets Thanks for the review! Could I trouble you to tag a release? Thanks again.

stanhu:
@markets Sorry to bother you again. Do you think you could cut a new release?

markets:
@stanhu I'm preparing a new release 👉🏼 #444

PR #442, opened 2023-11-28
Parse arguments on args= method

francktrouillez:
Relates to this [issue](https://github.com/sidekiq-cron/sidekiq-cron/issues/441)

The idea is to parse the `args` when we assign it.
It overrides the current `args=` method added by `attr_accessor`.

### DIFF
--- a/test/unit/job_test.rb
+++ b/test/unit/job_test.rb
@@ -340,7 +340,7 @@ describe "Cron Job" do
       end
     end
 
-    describe 'with GlobalID::Identification args' do      
+    describe 'with GlobalID::Identification args' do
       before do
         @args.merge!(args: Person.new(1))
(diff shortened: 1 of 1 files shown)


lib/sidekiq/cron/job.rb
```diff
@@ -294,7 +294,7 @@ def initialize input_args = {}
 
         # Get right arguments for job.
         @symbolize_args = args["symbolize_args"] == true || ("#{args["symbolize_args"]}" =~ (/^(true|t|yes|y|1)$/i)) == 0 || false
-        @args = args["args"].nil? ? [] : parse_args( args["args"] )
+        @args = parse_args(args["args"])
```
francktrouillez, line 297: :information_source: The check was unnecessary because `parse_args(nil)` is already `[]`

markets (approved):

PR #445, opened 2023-12-13
feat: human readable UI

Linell:
This changes the UI to be more human readable, using the [cronex](https://github.com/alpinweis/cronex) library to generate a human version of each job's cron. It also sets the `title` property of the job's name to the description, so that you're able to hover over the job and see the description without clicking into the show page.

Screen Shot 2023-12-13 at 8 21 40 AM


markets (approved):
Nice one! Thanks @Linell

PR #448, opened 2024-01-11
 Fetch queue name from ActiveJob class

### LINKED ISSUE #437: [ActiveJob] Cannot use `queue_as` for select queue, only sidekiq options are supported (first run)
Hi :wave: 

looks like sidekiq-cron works "okayish" with ActiveJob. It detects and runs jobs:)

There are 2 problems I'm running into

1. [here](https://github.com/sidekiq-cron/sidekiq-cron/blob/78eb493adbfeb908861ec9b3a76e5e7a01adf339/lib/sidekiq/cron/job.rb#L322) only sidekiq options are fetched on init, if sidekiq options change later on, this will not be fetched
2. When some of the options change, for example queue, it will be **only** overwritten if passed by args [see here](https://github.com/sidekiq-cron/sidekiq-cron/blob/78eb493adbfeb908861ec9b3a76e5e7a01adf339/lib/sidekiq/cron/job.rb#L334).

No code here supports ActiveJob configuration.

Is this more of a bug and PRs to get better ActiveJob is welcome or there's something wrong I'm doing?

Repro is very minimal

```yml
  foobar:
    cron: "*/5 * * * *" 
    class: "FooBar"
```

```rb
  class FooBar < ApplicationJob
    queue_as :low

    def perform
```

arathunku:
ActiveJob queue_as configuration is now used to update Job's queue, previously only Sidekiq::Worker jobs class options were fetched.

As in case with Sidekiq::Workers, queue option in config always takes precedence over queue_as in ActiveJob class.

Fixes #437

### DIFF
--- a/test/unit/job_test.rb
+++ b/test/unit/job_test.rb
@@ -273,6 +273,18 @@ describe "Cron Job" do
       assert enqueue_args[-1].is_a?(Float)
       assert enqueue_args[-1].between?(Time.now.to_f - 1, Time.now.to_f)
     end
+
+    describe "when job is Active Job worker class" do
+      it "be initialized with default attributes" do
+        job = Sidekiq::Cron::Job.new('klass' => 'ActiveJobCronTestClass')
+        assert_equal job.message, {"queue"=>"default", "class"=>"ActiveJobCronTestClass", "args"=>[]}
+      end
+
+      it "be initialized with class specified attributes" do
+        job = Sidekiq::Cron::Job.new('class' => 'ActiveJobCronTestClassWithQueue')
+        assert_equal job.message, {"queue"=>:super, "class"=>"ActiveJobCronTestClassWithQueue", "args"=>[]}
+      end
+    end
   end
 
   describe "cron test" do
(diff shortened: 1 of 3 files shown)


test/unit/job_test.rb
```diff
@@ -643,7 +655,7 @@
 
         describe 'with non sidekiq job class' do
           before do
-            @job = Sidekiq::Cron::Job.new(@args.merge(klass: 'CronTestClass'))
+            @job = Sidekiq::Cron::Job.new(@args.merge(klass: 'ActiveJobCronTestClass'))
```
arathunku, line 658: This is not clearly visible on diff, but this is in `active_job == true` block thus I'd say it's safe to assume any klass at least mimics ActiveJob interface. This may be a breaking change if someone's using  'active_job == true` in config with non-ActiveJob class because `queue_name` on class might not be available :cry:

markets (approved):
Thanks @arathunku!

arathunku:
Woa, that was fast @markets! Thank you!!!

PR #449, opened 2024-01-19
Add warning log when 'symbolize_args: true' is used without 'active_job: true'

francktrouillez:
Relates to this [issue](https://github.com/sidekiq-cron/sidekiq-cron/issues/443)

- Logs a warning when we use `symbolize_args: true` without using `active_job: true`.
- Adds this specifity in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,6 +172,8 @@ class HardWorker
 end
 ```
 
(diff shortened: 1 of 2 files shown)


markets (approved):
Thanks @francktrouillez! Probably enough for now.

PR #459, opened 2024-04-05
Add strict cron parsing mode

kevinrobell-st:
This adds a configuration option for stricter cron string parsing.

Specifically, I wanted to avoid the possibility of a natural language cron string being parsed into multiple cron lines and then everything but the first cron line getting ignored. This is the default when using `Fugit.do_parse_cron`.

```rb
irb(main):005:0> Fugit.do_parse("every day at 3:15 and 4:30")
=> #<Fugit::Cron:0x000000011100e170 @original="15 3 * * *", @cron_s=nil, @day_and=nil, @seconds=[0], @minutes=[15], @hours=[3], @monthdays=nil, @months=nil, @weekdays=nil, @zone=nil, @timezone=nil>
irb(main):006:0> Fugit.do_parse("every day at 3:15 and 4:30", :multi => :fail)
...
ArgumentError (multiple crons in "every day at 3:15 and 4:30" - {:monthday=>(slot :monthday "*" {:weak=>true})})
```

We use the `:multi => :fail` option to throw an error in those situations.

Closes: https://github.com/sidekiq-cron/sidekiq-cron/issues/450

CC: @jmettraux

### DIFF
--- a/lib/sidekiq/cron.rb
+++ b/lib/sidekiq/cron.rb
@@ -21,6 +21,19 @@ module Sidekiq
 
       def initialize
         @default_namespace = 'default'
+        @strict_cron_parsing = false
+      end
+
+      # Throws an error when a natural language cron string would get
+      # parsed into multiple cron lines. By default the `fugit` gem is
+      # permissive when parsing natural language cron strings. Only the first
+      # cron line is used in that case and all other ones are ignored.
+      def strict_cron_parsing!
+        @strict_cron_parsing = true
+      end
+
+      def strict_cron_parsing?
+        @strict_cron_parsing
       end
     end
   end
--- a/lib/sidekiq/cron.rb
+++ b/lib/sidekiq/cron.rb
@@ -19,21 +19,23 @@ module Sidekiq
       # The default namespace is used when no namespace is specified.
       attr_accessor :default_namespace
 
+      # The parsing mode when using the natural language cron syntax from the `fugit` gem.
+      #
+      # :single -- use the first parsed cron line and ignore the rest (default)
+      # :strict -- raise an error if multiple cron lines are parsed from one string
+      attr_reader :natural_cron_parsing_mode
+
       def initialize
         @default_namespace = 'default'
-        @strict_cron_parsing = false
+        @natural_cron_parsing_mode = :single
       end
 
-      # Throws an error when a natural language cron string would get
-      # parsed into multiple cron lines. By default the `fugit` gem is
-      # permissive when parsing natural language cron strings. Only the first
-      # cron line is used in that case and all other ones are ignored.
-      def strict_cron_parsing!
-        @strict_cron_parsing = true
-      end
+      def natural_cron_parsing_mode=(mode)
+        unless %i[single strict].include?(mode)
+          raise ArgumentError, "Unknown natural cron parsing mode: #{mode.inspect}"
+        end
 
-      def strict_cron_parsing?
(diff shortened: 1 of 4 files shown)


markets:
Hi @kevinrobell-st 

I just took a quick look and it looks fine 👌.

lib/sidekiq/cron.rb
```diff
@@ -21,6 +21,19 @@ class Configuration
...
       def initialize
         @default_namespace = 'default'
+        @strict_cron_parsing = false
+      end
+
+      # Throws an error when a natural language cron string would get
+      # parsed into multiple cron lines. By default the `fugit` gem is
+      # permissive when parsing natural language cron strings. Only the first
+      # cron line is used in that case and all other ones are ignored.
+      def strict_cron_parsing!
```
markets, line 31: Maybe we can use an attr_accessor instead of a bang method 🤔?

jmettraux:
Hello @kevinrobell-st and @markets,

thanks for pinging me. It looks good to me.

(now it opens the question: since `"every day at 3:15 and 4:30"` when not `strict_cron_parsing == false` returns two crons, should Sideqik-Cron embrace this multiplicity, call fugit with `multi: true` and... Maybe `Sidekiq::Cron.configuration.natural_cron_parsing` should be an attribute that can yield `:strict`, `:single` (default, current), or `:multi`, a mode that can be implemented later on, should a real need arise...)

lib/sidekiq/cron.rb
kevinrobell-st: I have no strong preference between the two. Whichever you prefer.

kevinrobell-st:
> (now it opens the question: since `"every day at 3:15 and 4:30"` when not `strict_cron_parsing == false` returns two crons, should Sideqik-Cron embrace this multiplicity, call fugit with `multi: true` and... Maybe `Sidekiq::Cron.configuration.natural_cron_parsing` should be an attribute that can yield `:strict`, `:single` (default, current), or `:multi`, a mode that can be implemented later on, should a real need arise...)

That's an interesting suggestion. I'm not too familiar with the internals of Sidekiq Cron so I don't know how big of a change that would be. Ideally those multiple cron lines would still be tied to a single cron job. I've definitely run into a few times where that would have been useful including the schedule that brought this issue to my attention in the first place.

lib/sidekiq/cron.rb
markets: I think it will be more consistent with other option flags that may arise in the future.

markets:
Thanks @jmettraux 🙌 

Really interesting proposal. I think we can merge this one for now, and we'll see how this feature evolves.

kevinrobell-st:
Maybe the best way to go about this would be to make the configuration option line up with that described in https://github.com/sidekiq-cron/sidekiq-cron/pull/459#issuecomment-2040671426 so that things don't have to be changed that much should more parsing get added in the future. I'll update the PR to do that.

markets (approved):
Nice job @kevinrobell-st 👌

I would like to hear @jmettraux final feedback on this one, as his opinions help us a lot to take good decisions regarding this part. Then, I'll probably cut a v2 beta release, we have interesting changes in the cooking: https://github.com/sidekiq-cron/sidekiq-cron/compare/v1.12.0...master

jmettraux:
Hello @markets and @kevinrobell-st, thanks for keeping me in the loop.

Looks all good to me! Great job @kevinrobell-st !

PR #460, opened 2024-04-15
Move class attributes and initialize method at the beginning of the file

markets:
This always bothered me a little 😅 I think it's more natural to have `attr_accessor`s and the `initialize` method at the beginning of the class.


pboling:
Agreed. I love the class hierarchy rubocop for this, though i have now forgotten (again) where it comes from.

PR #465, opened 2024-04-21
Add ability to configure the past scheduled time threshold

### LINKED ISSUE #463: Ability to configure the past schedule interval
We are going to use this gem in our application, instead of our custom solution. 

One problem we had and this gem does not fully solve is that if sidekiq is restarting while some schedule is about to run, we may loose that schedule. For example, if we have some job to run at `15:00` and deployment is happening at `14:57` (and sidekiq is restarted) and sidekiq is back up to live at 15:03, then we missed that job to be scheduled.

This can be solved by zero downtime deployments or smaller more frequent jobs or some refactoring, but that is not currently achievable.

If the `60` seconds from https://github.com/sidekiq-cron/sidekiq-cron/blob/abdb335a796d3cc3841b8029ee8da26f0bde08e6/lib/sidekiq/cron/job.rb#L668-L672 will be configurable, then this will save us and a bit older missed schedules will be considered.

Are you open for this feature? Would be glad to provide a PR.

fatkodima:
Closes #463.

Naming is hard, so probably people can suggest a better name for this configuration option.

I also refactored a bit the relevant methods to avoid double negation and be simpler to understand.

### DIFF
--- a/README.md
+++ b/README.md
@@ -451,6 +451,16 @@ Sidekiq-Cron is checking jobs to be enqueued every 30s by default, you can chang
 Sidekiq::Options[:cron_poll_interval] = 10
 ```
 
+When sidekiq (and sidekiq-cron) is not used in zero-downtime deployments, after the deployment is done sidekiq-cron starts to catch up. It will consider older jobs that missed their schedules during that time. By default, only jobs that should have started less than 1 minute ago are considered. This is problematic for some jobs, e.g., jobs that run once a day. If on average sidekiq is shut down for 10 minutes during deployments, you can configure sidekiq-cron to consider jobs that were about to be scheduled during that time:
+
+```ruby
+# config/initializers/sidekiq-cron.rb
+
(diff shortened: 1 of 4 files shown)


markets:
Seems fine @fatkodima ✅

Let's leave it open some days, so we can think a bit on the naming.

kevinrobell-st:
Maybe something like "reschedule grace period" would make more sense in terms of naming.

markets:
Like it @kevinrobell-st! What's your thoughts @fatkodima?

fatkodima:
Like it! Switched to that name, thank you ❤️

markets (approved):

markets:
Thanks @fatkodima 🙌

markets:
Hopefully soon! In the meantime, you can point the gem to GitHub `master` branch: 

```ruby
gem 'sidekiq-cron', github: 'sidekiq-cron/sidekiq-cron'
```

PR #475, opened 2024-08-15
Require at least fugit >= 1.11.1

jmettraux:
https://github.com/floraison/fugit/issues/104

Prevent Fugit::Nat.parse choking on large input, limit at 256 chars


markets (approved):
Thanks @jmettraux

PR #476, opened 2024-08-29
Fix: according to documentation, `.create` should return true/false

delphaber:
Related to https://github.com/sidekiq-cron/sidekiq-cron/issues/471

@markets I fixed it in the most easy way at the moment: by reaching the end of the method we can "suppose" that everything worked as expected and it can return `true`.

Actually, in the `Sidekiq.redis` block we are calling some redis commands (`SADD`, `HSET`, `ZADD`) but the code is not checking their return values. It's unlikely that those commands fail, unless a connection error occurs (but that would raise an expetion, I suppose!). So this might be good enough. What do you think?

### DIFF
--- a/lib/sidekiq/cron/job.rb
+++ b/lib/sidekiq/cron/job.rb
@@ -490,6 +490,8 @@ module Sidekiq
           conn.zadd(job_enqueued_key, time.to_f.to_s, formatted_last_time(time).to_s) unless exists == true || exists == 1
         end
         Sidekiq.logger.info { "Cron Jobs - added job with name #{@name} in the namespace #{@namespace}" }
+
+        true
       end
 
       def save_last_enqueue_time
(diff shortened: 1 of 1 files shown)


delphaber:
So, if `exists` is `true` or `1` the `#save` method should return false and the log line should not be printed. Have I understood it correctly?

sandip-mane:
@delphaber you're right. if `exists` is `true/1` we should return true without printing the log.

delphaber:
OK, I think now it does what we said earlier

markets:
Thanks @delphaber! @sandip-mane Could you please 🙏🏼 also perform a code review for this one?

markets (approved):

sandip-mane (approved):
Looks good to me 🔥

Issue #474, opened 2024-08-08
Add Pagination to Cron Job Listing Page for Improved Usability


satyakampandya:
Currently, the cron job listing page in the Sidekiq-Cron web UI displays all cron jobs in a single view. This can become cumbersome and inefficient when managing a large number of jobs, as users must scroll extensively to find or review specific entries. Adding pagination to this page would enhance the user experience by making navigation more manageable and the interface more responsive.

markets:
Closing for now, please read the discussion in #477 for further information.

Issue #468, opened 2024-05-08
Sidekiq 7.3 compatibility


mperham:
Sidekiq 7.3 will remove support for inline styling and scripts. This means you'll need to move your dynamic styling into a separate CSS file and flag it with the current nonce.

```html
<link href="<%= root_path %>stylesheets/sidekiq-cron.css" media="screen" rel="stylesheet" type="text/css" nonce="<%= csp_nonce %>" />
```

Where `sidekiq-cron.css` should go into a directory which you have added to Sidekiq::Web's asset paths. I don't have an API for this yet -- TBD.

See sidekiq/sidekiq#6270

markets:
Thanks for the info @mperham! Will take a look asap this new API is defined. 

Just a question for now, I understand this will only work with Sidekiq 7.3+ right? That means, that we'll need to somehow maintain 2 ways (pre and post 7.3) of "injecting" our assets to Sidekiq Web UI?

mperham:
Yep, unfortunately that is correct today. I'm open to suggestions for how to make this easier. This change is designed to remove an entire style of attack on the Web UI (XSS) but requires all /sidekiq pages to be compliant or they will break. There's 4-5 popular Sidekiq extensions so I'm hoping this will be doable but some of them have significant UI.

markets:
Hello @mperham, 

I reviewed all our front-end part for the Web UI, and we are not using any external js or css, just a couple of dynamically-generated inline styles: https://github.com/search?q=repo%3Asidekiq-cron%2Fsidekiq-cron%20style&type=code. Do you think this is gonna be a problem?

mperham:
We're going to allow inline styles for 7.3 but they'll need to move to static classes + predefined javascript for 8.0. Right now I'm writing an example extension which you can use as a reference for how to do things the "correct" way.

markets:
Ok, thanks Mike! Please, share the example once you have a chance, I'll try refactor our inline styles asap, so we'll prepared for compatibility with the future Sidekiq release.

mperham:
Here's the new registration API and example extension. Let me know what you think, everything is still malleable so feedback would be appreciated. I still need to add a test.

https://github.com/sidekiq/sidekiq/pull/6292

satyakampandya:
[...]

PR #477, opened 2024-09-02
Feature: Pagination Feature for Cron Jobs

satyakampandya:
# Pagination Feature for Cron Jobs

This PR updates the Redis commands used in the application to manage cron jobs from SADD (set add) to ZADD (sorted set add). The primary purpose of this change is to leverage the ordering capabilities of Redis sorted sets, which allows for better control over the data order and facilitates easier pagination.

## Data Structure Update:
- Changed from using Redis sets (SADD) to sorted sets (ZADD) to store cron jobs.
- Each cron job is now stored with an associated score that represents its creation time, which allows for sorted retrieval.

## Modification of Adding Members:
- The command SADD jobs_key member is replaced with ZADD jobs_key score member.
- Example change:
```ruby
# Old code using SADD
conn.sadd self.class.jobs_key(@namespace), [redis_key]

# New code using ZADD with time as the score
time = Time.utc.now
conn.zadd self.class.jobs_key(@namespace), time.to_f.to_s, [redis_key]
```

## Pagination Implementation:
- Previously, pagination might have been manually handled or based on unordered sets.
- With ZADD, we can now efficiently paginate results using the scores in the sorted set.
- Example retrieval change:
```ruby
# Old code using SMEMBERS
conn.smembers(jobs_key(namespace))

# New code using ZRANGE with score-based pagination
conn.zrange(jobs_key(namespace), start, stop)
```

## Test Cases Updated:
- Updated unit tests to reflect the change from sets to sorted sets.
- Added new test cases to validate the ordering and pagination of cron jobs.

## Benefits of This Change
- **Improved Data Ordering:** Using ZADD allows us to store data in a sorted order, which is beneficial for retrieving data in a specific sequence without additional sorting logic.
- **Efficient Pagination:** The ordered nature of sorted sets enables straightforward pagination using Redis commands like ZRANGE and ZREVRANGE.
- **Performance Improvements:** By leveraging Redis’s native sorted set operations, we reduce the complexity and potentially increase the performance of data retrieval operations.

## Related Issues
- Feature Request: #474 

## Note
This is my first PR to an open-source project, and I’m excited to contribute! I welcome any feedback or suggestions for improvement. If there are better approaches or practices I could follow, I’m eager to learn and make the necessary adjustments. Thank you for your support and guidance!

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,6 +184,9 @@ Sidekiq::Cron::Job.all 'Foo'
 Sidekiq::Cron::Job.all '*'
 #=> [#<Sidekiq::Cron::Job ...>]
 
+# If you'd like to paginate jobs, you can pass offset and limit
+Sidekiq::Cron::Job.all(offset: 0, limit: 9)
+
 job = Sidekiq::Cron::Job.find('Hard worker - every 5min', 'Foo').first
 job.destroy
 # INFO: Cron Jobs - deleted job with name Hard worker - every 5min from namespace Foo
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
 
 🎬 [Introduction video about Sidekiq-Cron by Drifting Ruby](https://www.driftingruby.com/episodes/periodic-tasks-with-sidekiq-cron)
 
-Sidekiq-Cron runs a thread alongside Sidekiq workers to schedule jobs at specified times (using cron notation `* * * * *` parsed by [Fugit](https://github.com/floraison/fugit)).
+Sidekiq-Cron runs a thread alongside Sidekiq workers to schedule jobs at specified times (using cron notation `* * * * *` or natural language, powered by [Fugit](https://github.com/floraison/fugit)).
 
 Checks for new jobs to schedule every 30 seconds and doesn't schedule the same job multiple times when more than one Sidekiq process is running.
 
@@ -56,9 +56,9 @@ gem "sidekiq-cron"
   'queue' => 'name of queue',
   'args' => '[Array or Hash] of arguments which will be passed to perform method',
   'date_as_argument' => true, # add the time of execution as last argument of the perform method
-  'active_job' => true,  # enqueue job through Rails 4.2+ Active Job interface
-  'queue_name_prefix' => 'prefix', # Rails 4.2+ Active Job queue with prefix
-  'queue_name_delimiter' => '.', # Rails 4.2+ Active Job queue with custom delimiter (default: '_')
+  'active_job' => true,  # enqueue job through Active Job interface
+  'queue_name_prefix' => 'prefix', # Active Job queue with prefix
+  'queue_name_delimiter' => '.', # Active Job queue with custom delimiter (default: '_')
   'description' => 'A sentence describing what work this job performs'
   'status' => 'disabled' # default: enabled
 }
@@ -79,7 +79,7 @@ like this `'0 22 * * 1-5 America/Chicago'`.
 
 #### Natural-language formats
 
-Since sidekiq-cron `v1.7.0`, you can use the natural-language formats supported by Fugit, such as:
+Since Sidekiq-Cron `v1.7.0`, you can use the natural-language formats supported by Fugit, such as:
 
 ```rb
 "every day at five" # => '0 5 * * *'
@@ -122,7 +122,7 @@ Ex. `every day at 3:15 and 4:30`
 
 #### Second-precision (sub-minute) cronlines
 
-In addition to the standard 5-parameter cronline format, sidekiq-cron supports scheduling jobs with second-precision using a modified 6-parameter cronline format:
+In addition to the standard 5-parameter cronline format, Sidekiq-Cron supports scheduling jobs with second-precision using a modified 6-parameter cronline format:
 
 `Seconds Minutes Hours Days Months DayOfWeek`
 
@@ -225,7 +225,7 @@ class ExampleJob < ActiveJob::Base
 end
 ```
 
-For Active jobs you can use `symbolize_args: true` in `Sidekiq::Cron::Job.create` or in Hash configuration,
+For Active Job you can use `symbolize_args: true` in `Sidekiq::Cron::Job.create` or in Hash configuration,
 which will ensure that arguments you are passing to it will be symbolized when passed back to `perform` method in worker.
 
 ### Adding Cron jobs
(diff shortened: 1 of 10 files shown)


satyakampandya:
### Note on Pagination with * Namespace
Currently, pagination is not supported when the namespace is set to * due to the complexity of managing pagination across multiple namespaces within a loop. One possible solution could be to add an entirely new sorted set that combines all namespaces, which would simplify the pagination logic. However, I'm unsure if this approach is ideal or if it aligns with the overall design principles of the project.

I've added a TODO comment in the code regarding this limitation. If anyone has thoughts on this approach or other suggestions for implementing pagination in this scenario, please share your ideas. I’m open to feedback and improvements!

markets:
Hi @satyakampandya 👋🏼 

First of all, thank you so much for such detailed explanations in this very well formatted PR 👏🏼. Good job!

About the feature: I'm not against adding something like this in the gem, but the code changes seems a bit sensible? I'm not currently very familiar with the Redis internals, so not sure how these changes will impact people updating to this eventual future release. Maintaining complex open-source libraries in the long term is hard, so every time we add a new feature, we are adding more maintenance for the future. 

To be honest, I personally don't need this feature and don't want to maintain more complexity in this library sorry 🙏🏼 ... but if a lot of people wants this, I'm also ok, but I'd like more people to perform code review, give feedback and testing in their apps.

P.S. For the * Namespace thing, I think it will be ok to just raise an exception with something like "You should set a namespace to use the pagination".

markets:
Hello @satyakampandya and thank you again for all the detailed comments 👍🏼

I'm personally not interested in this feature and to deal with more complexity (plus the migration script). The next release, which will be `v2.0.0`, will include another big change: https://github.com/sidekiq-cron/sidekiq-cron/pull/268. I think that performing 2 data migrations during an upgrade is probably too much and potentially bad things can happen (you know, some times shit happens). So, sorry 🙏🏼 but I'm personally not interested in merging and maintaining this feature, at least for now... but if other maintainers want to take the responsibility, that's fine for me too.

satyakampandya:
@markets I completely understand your concerns about handling multiple data migrations in the upcoming release. I appreciate the consideration and respect your decision. If there are any other maintainers who might be interested in taking this forward, I’d be happy to discuss it with them.

Additionally, if there are other features or issues where you need help, I’d be more than happy to contribute.

markets:
Thank you so much for offering your help @satyakampandya! Really appreciated 👌🏼 

If you would like to help with this one #468, I would love to merge it asap.

satyakampandya:
Hello @markets 
Sure, I'll have a look at it.

freemanoid:
I like this feature overall. It should make the UI cleaner. I have a couple of ideas:

1. Pagination without a way to find a particular schedule by name complicates any interaction with the UI. So having a search form that would filter schedules by name would be helpful.
2. I don't think it's a good idea to change redis data structure just for the web UI pagination. We're not talking about tens of thousands of schedules here. A simpler solution would be to avoid changing the data structure and just use `#smembers` and sort the results in ruby. Also, it should be fine to pass an entire array of schedules to the frontend, so the search can be implemented on the frontend without making any additional queries to the backend.

satyakampandya:
@markets, I’ve reverted the sorted sets implementation and switched to using the [SORT](https://redis.io/docs/latest/commands/sort/) method as follows:
```ruby
conn.sort(jobs_key(namespace), 'LIMIT', offset, limit, 'ALPHA')
```
This change works successfully on my local setup and passes all the tests. However, in the CI tests for this PR, I’m encountering the following error:
```
ArgumentError: wrong number of arguments (given 5, expected 1)
    vendor/bundle/ruby/3.3.0/gems/redis-4.8.1/lib/redis/commands/keys.rb:407:in `sort'
```
According to the method definition [here](https://github.com/redis/redis-rb/blob/v4.8.1/lib/redis/commands/keys.rb#L407), it should be used like this:
```ruby
conn.sort(jobs_key(namespace), order: "alpha", limit: [offset, limit])
```
This approach doesn’t work locally. Do you have any insights on this?

freemanoid:
> This approach doesn’t work locally. Do you have any insights on this?

Specs are failing on sidekiq 6. Sidekiq 7 has moved to a different redis client, that is probably the root cause.

satyakampandya:
Thank you @freemanoid for pointing out the issue. 

@markets @freemanoid In this implementation, we preserve the existing Redis data structure without adding any complexity. The changes are simple, utilizing the SORT method rather than altering data structures, which should make ongoing maintenance easier.

If you’re open to reconsidering this PR for the upcoming release, I believe it will improve the UI experience for users while keeping the Redis structure intact.

freemanoid:
> If you’re open to reconsidering this PR for the upcoming release, I believe it will improve the UI experience for users while keeping the Redis structure intact.

IMO adding pagination without a way to filter jobs by name is worse than what we have currently. I never go to the cron UI to "browse" the jobs. I need to find a particular job by a substring. Currently, I search on a page using a browser search.

kevinrobell-st:
How does this new pagination approach handle disabled schedules? Currently disabled cron schedules are shown in a block at the bottom of the cron page in the Sidekiq web UI.

markets:
The new approach, without changing the data structures, is much more feasible to be eventually merged. But I'm still a bit doubtful about this feature in general. Good points raised above by @freemanoid @kevinrobell-st.

satyakampandya:
> IMO adding pagination without a way to filter jobs by name is worse than what we have currently. I never go to the cron UI to "browse" the jobs. I need to find a particular job by a substring. Currently, I search on a page using a browser search.
@freemanoid, I understand your concern about filtering jobs by name, but I believe pagination is still valuable, especially since we already have a namespace filter.

> How does this new pagination approach handle disabled schedules? Currently disabled cron schedules are shown in a block at the bottom of the cron page in the Sidekiq web UI.
@kevinrobell-st, instead of moving disabled cron schedules to the bottom, we can update the UI to display them in their default positions while clearly marking them as disabled. This approach maintains the overall order and visibility.

kevinrobell-st:
I think that approach will make it harder to find any individual job or group of disabled jobs. Having to potentially click through all of the pages to find a job seems like a frustrating user experience.

markets:
I agree with @kevinrobell-st... I think we will need too much changes in the code and extra complexity to support pagination in the admin panel (with no clear benefits between current solution). 

The current solution with namespaces can help to group jobs and you can also use the browser native search if you're looking for a specific job. At the end of the day, this is just like an internal page for developers... It's not worth it. I'd just close it.

satyakampandya:
At this point, I don't see an optimal solution for implementing the search functionality, so I suggest we go ahead and close this PR for now.

Issue #478, opened 2024-09-05
Deprecation Warning "Using the last argument as keyword parameters is deprecated"


jvanderen1:
In our codebase, we have Ruby 3 deprecation warnings enabled via:

```ruby
Warning[:deprecated] = true
```

When we run a cron job via `load_from_hash`:

```ruby
schedule_file = "config/schedule.yml"

if File.exist?(schedule_file) && Sidekiq.server?
  schedule_hash = YAML.load_file(schedule_file)
  Sidekiq::Cron::Job.load_from_hash schedule_hash
end
```

We always end up seeing a warning message:

```
/usr/local/bundle/gems/sidekiq-cron-1.10.1/lib/sidekiq/cron/job.rb:472: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
```

Is this a known issue with `sidekiq-cron` or has this been fixed in versions > 1.10.1?

markets:
Hello @jvanderen1 and thanks for reporting this warning. This is not tracked yet in any issue or PR, so feel free to send a patch! We'd be very happy to review it 🤝 

Btw, you can already remove this file now, as it will be automatically picked by the [gem itself](https://github.com/sidekiq-cron/sidekiq-cron/blob/master/lib/sidekiq/cron/schedule_loader.rb): https://github.com/sidekiq-cron/sidekiq-cron?tab=readme-ov-file#loading-jobs-from-schedule-file.

jvanderen1:
@markets Thank you for the suggestion! In the example above, I did omit some smaller details. For example, before we start any cronjobs, we actually delete any active one's first.

i.e.
```ruby
### First delete all jobs
Sidekiq::Cron::Job.all.each(&:destroy)

### Then load config/schedule.yml and restart
schedule_file = "config/schedule.yml"
if File.exist?(schedule_file) && Sidekiq.server?

  schedule_hash = YAML.load_file(schedule_file)
  Sidekiq::Cron::Job.load_from_hash schedule_hash
end
```

Is this necessary to do or will the gem automatically delete stale jobs when `require`d?

markets:
In the schedule loader we are using the bang suffixed methods:

https://github.com/sidekiq-cron/sidekiq-cron/blob/81268643b8fe4bb0a2197cb650b16a115a220c6f/lib/sidekiq/cron/schedule_loader.rb#L12

[...]

PR #479, opened 2024-09-05
Update How Redis Values Are Stored on `save`

jvanderen1:
Resolves https://github.com/sidekiq-cron/sidekiq-cron/issues/478

The current way of setting values is to use a hash. As noted in the issue attached, this format is marked as deprecated by Ruby 3. The following warning is thrown:

```
/usr/local/bundle/gems/sidekiq-cron-1.10.1/lib/sidekiq/cron/job.rb:472: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
```

This will instead pass an array to Redis. This is how [`hmset`](https://www.rubydoc.info/github/ezmobius/redis-rb/Redis:hmset) accepts arguments. Also, `hset` is preferred over `hmset` as Redis has deprecated the `hmset` command.

## Tested

I was able to get all tests to pass locally

I can also see the keys being set correctly via the `redis-cli`:

```
> HGET cron_job:foo_job description
"A temporary job to check I am alive"
```

### DIFF
--- a/lib/sidekiq/cron/job.rb
+++ b/lib/sidekiq/cron/job.rb
@@ -482,7 +482,7 @@ module Sidekiq
           conn.sadd self.class.jobs_key(@namespace), [redis_key]
 
           # Add information for this job!
-          conn.hset redis_key, to_hash.transform_values { |v| v || '' }.flatten
+          conn.hset redis_key, to_hash.transform_values! { |v| v || '' }.flatten
 
           # Add information about last time! - don't enque right after scheduler poller starts!
           time = Time.now.utc
--- a/lib/sidekiq/cron/job.rb
+++ b/lib/sidekiq/cron/job.rb
@@ -482,7 +482,7 @@ module Sidekiq
           conn.sadd self.class.jobs_key(@namespace), [redis_key]
 
           # Add information for this job!
-          conn.hset redis_key, to_hash.transform_values! { |v| v || '' }
+          conn.hset redis_key, *to_hash.transform_values { |v| v || '' }.flatten
 
           # Add information about last time! - don't enque right after scheduler poller starts!
           time = Time.now.utc
--- a/lib/sidekiq/cron/job.rb
+++ b/lib/sidekiq/cron/job.rb
@@ -482,7 +482,7 @@ module Sidekiq
           conn.sadd self.class.jobs_key(@namespace), [redis_key]
 
           # Add information for this job!
-          conn.hset redis_key, *to_hash.transform_values { |v| v || '' }.flatten
+          conn.hset redis_key, to_hash.transform_values { |v| v || '' }.flatten
 
           # Add information about last time! - don't enque right after scheduler poller starts!
           time = Time.now.utc


markets:
Thanks @jvanderen1! I think that makes sense ✅

But not sure why you link to the `hmset` docs, as we are using `hset` here 🤔. I'd like to merge it, but could you please clarify that?

jvanderen1:
@markets Apologies for the confusion! I found the Redis gem's documentation to be a little confusing myself.

## TLDR

The official Redis gem should update the documentation for `hset` to indicate support for multiple field values.

## Background of `hset` vs `hmset`

When looking at the documentation for [`hset`](https://www.rubydoc.info/gems/redis/3.1.0/Redis:hset), the method signature looks like:

```
#hset(key, field, value) ⇒ Boolean

Parameters:
  key (String)
  field (String)
  value (String)
```

However, the documentation for [`hmset`](https://www.rubydoc.info/gems/redis/3.1.0/Redis:hmset) indicates it's signature looks like:

```
#hmset(key, *attrs) ⇒ Object

Parameters:
  key (String)
  attrs (Array<String>) — array of fields and values
```

Seeing `hmset` looks more inline with setting multiple field values, I opted to try that. However, the hmset method gave me this deprecation warning when I used it:

```
[sidekiq#5788] Redis has deprecated the `hmset`command, called at [".../ruby/2.7.8/lib/ruby/gems/2.7.0/gems/sidekiq-cron-1.10.1/lib/sidekiq/cron/job.rb:472:in `block in save'"]
```

Looking at the official [Redis documentation](https://redis.io/docs/latest/commands/hmset/):

> As of Redis version 4.0.0, [HMSET] is regarded as deprecated.
>
> It can be replaced by [HSET](https://redis.io/docs/latest/commands/hset/) with multiple field-value pairs when migrating or writing new code.

markets:
Ok! Thanks for clarifying @jvanderen1. 

Let me add a couple of notes:
1) Btw, not sure how this PR solves the warning described in #478. They seem like different things no?
2) The current `transform_values!` saves some memory, as it doesn't create a new hash every time... but probably that's not a hot-spot in this code base.

jvanderen1:
@markets

> Btw, not sure how this PR solves the warning described in https://github.com/sidekiq-cron/sidekiq-cron/issues/478. They seem like different things no? Sorry if I'm confused 🙏🏼 just want to make sure before I hit the "merge" button.

I updated the description with the warning I see before these changes:

```
/usr/local/bundle/gems/sidekiq-cron-1.10.1/lib/sidekiq/cron/job.rb:472: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
```

When a hash is passed into a method which can take keyword arguments technically behaves different in Ruby 2 vs Ruby 3. This changes the shape from a hash to an array, therefore ensuring that a regular argument is passed instead of a keyword argument.

> The current transform_values! saves some memory, as it doesn't create a new hash every time... but probably that's not a hot-spot in this code base.

To reduce the scope of changes, I will revert back to `transform_keys!`. The assumption was that `transform_keys!` may take longer than `transform_keys`, but that may be incorrect.

markets:
Thanks again @jvanderen1 🚀

markets (approved):

PR #480, opened 2024-09-11
Add compatibility with Sidekiq 7.3+ and remove inline styles

satyakampandya:
#### Support for Sidekiq 7.3+:
- Updated the code to use the new API for web extensions in Sidekiq 7.3 and above.
- Added checks to ensure it still works with older Sidekiq versions.

#### Remove Inline Styles:
- Removed all inline CSS styles from the views to make the code cleaner.
- Used existing CSS classes instead, so the design stays consistent without extra styles.

#### `cron.erb` and `cron_show.erb`:
- Removed inline styles and used existing CSS classes.
- Simplified some parts of the view to make the code easier to maintain.

#### `web.rb`:
- Added logic to register the web extension using the new Sidekiq 7.3+ API.
- Included a fallback for older Sidekiq versions to keep everything working.

### References:
- https://github.com/sidekiq/sidekiq/pull/6292
- https://github.com/sidekiq-cron/sidekiq-cron/issues/468#issuecomment-2341467667

### DIFF
--- a/lib/sidekiq/cron/views/cron.erb
+++ b/lib/sidekiq/cron/views/cron.erb
@@ -31,7 +31,7 @@
 <div class='row'>
   <div class="col-sm-12 summary_bar">
     <ul class="list-unstyled summary row">
-      <% @namespaces.sort_by { |namespace| namespace[:name] } .each do |namespace| %>
+      <% @namespaces.sort_by { |namespace| namespace[:name] }.each do |namespace| %>
         <li class="col-sm-1">
           <a href="<%= root_path %>cron/namespaces/<%= namespace[:name] %>">
             <span class="count"><%= namespace[:count] %></span>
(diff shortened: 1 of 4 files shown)


markets:
Nice job @satyakampandya, I'll take a deeper look during the following days but looks good ✅

Out of curiosity, did you try this branch in any Rails app? We don't have UI tests, so currently this needs some manual testing 👀

satyakampandya:
@markets I've tested this against several Sidekiq versions and found that it works well with versions up to Sidekiq v6.5.0. However, I encountered issues with versions below 6.5.0, even when using sidekiq-cron's master branch.

Given this, I think it makes sense to update the sidekiq dependency to require at least version 6.5.0. Here's the change I suggest adding to the `sidekiq-cron.gemspec`:

```ruby
s.add_dependency("sidekiq", ">= 6.5.0")
```

markets (approved):
Thanks @satyakampandya!

PR #482, opened 2024-09-12
Add confirmation for enqueuing disabled cron jobs and refactor job name escaping

satyakampandya:
This PR includes two improvements to the `cron.erb` view:

- **Add Confirmation for Disabled Jobs**: Ensures that when attempting to enqueue a disabled cron job, a confirmation dialog (`data-confirm`) is displayed. This brings consistency in behaviour, as the same confirmation is already shown for enabled jobs.

- **Refactor Job Name Escaping**: Replaces repeated occurrences of `CGI.escape(job.name).gsub('+', '%20')` with a single variable `escaped_job_name`, reducing redundancy and improving code readability.

These changes should improve the user experience by preventing accidental enqueuing of disabled jobs and simplify the codebase.

#### References:
- Closes https://github.com/sidekiq-cron/sidekiq-cron/issues/481

### DIFF
--- a/lib/sidekiq/cron/views/cron.erb
+++ b/lib/sidekiq/cron/views/cron.erb
@@ -59,10 +59,11 @@
     <tbody>
       <% @cron_jobs.sort{ |a,b| a.sort_name <=> b.sort_name }.each do |job| %>
         <% klass = (job.status == 'disabled') ? 'bg-danger text-muted' : '' %>
+        <% escaped_job_name = CGI.escape(job.name).gsub('+', '%20') %>
         <tr>
           <td class="<%= klass %>"><%= t job.status %></td>
           <td class="<%= klass %>">
-            <a href="<%= root_path %>cron/namespaces/<%= @current_namespace %>/jobs/<%= CGI.escape(job.name).gsub('+', '%20') %>" title="<%= job.description %>">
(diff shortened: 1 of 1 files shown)


markets:
@satyakampandya Maybe we can also try to fix this one? https://github.com/sidekiq-cron/sidekiq-cron/issues/452. 

I think it can be related to this "Refactor Job Name Escaping" thing.

satyakampandya:
Sure @markets 
Let me try to reproduce it and I'll add the fix in this PR.

satyakampandya:
Hi @markets,

I noticed that the namespace seems to have the same issue with handling special characters like forward slashes. After trying different approaches, the only solution that worked for both the job name and namespace was to encode them using `Base64.urlsafe_encode64(job.name)` and then decode them with `@job_name = Base64.urlsafe_decode64(route_params[:name])`.

This approach ensures the URLs remain intact, and special characters no longer break the routing. I’d love to hear your opinion on this approach—if you approve, I can go ahead and prepare the fix. 

Since this change would require updates in multiple places, I prefer to make it in a separate PR to keep things organised.

markets:
Agree! Then I think it should be fixed in another PR (low prio), seems out of the scope for this one.

markets (approved):

PR #484, opened 2024-09-14
Restrict namespace from using '*' and fix job actions for wildcard namespaces

satyakampandya:
This PR introduces a validation that disallows `*` as a valid namespace in sidekiq-cron jobs. It also fixes an issue where individual job actions like "Enqueue", "Delete", "Enable", and "Disable" fail when all jobs are listed using the '*' wildcard. The job's actual namespace is now correctly referenced for these actions. A test case has been added to ensure `*` is considered an invalid namespace.

Fixes: https://github.com/sidekiq-cron/sidekiq-cron/issues/483

### DIFF
--- a/lib/sidekiq/cron/views/cron.erb
+++ b/lib/sidekiq/cron/views/cron.erb
@@ -63,7 +63,7 @@
         <tr>
           <td class="<%= klass %>"><%= t job.status %></td>
           <td class="<%= klass %>">
-            <a href="<%= root_path %>cron/namespaces/<%= @current_namespace %>/jobs/<%= escaped_job_name %>" title="<%= job.description %>">
+            <a href="<%= root_path %>cron/namespaces/<%= job.namespace %>/jobs/<%= escaped_job_name %>" title="<%= job.description %>">
               <b class="<%= klass %>"><%= job.name %></b>
             </a>
             <br/>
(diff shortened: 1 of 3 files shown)


markets:
@zedtux What do you think about these changes? Would like to hear your thoughts 👀

lib/sidekiq/cron/views/cron.erb
```diff
@@ -63,7 +63,7 @@
         <tr>
           <td class="<%= klass %>"><%= t job.status %></td>
           <td class="<%= klass %>">
-            <a href="<%= root_path %>cron/namespaces/<%= @current_namespace %>/jobs/<%= escaped_job_name %>" title="<%= job.description %>">
+            <a href="<%= root_path %>cron/namespaces/<%= job.namespace %>/jobs/<%= escaped_job_name %>" title="<%= job.description %>">
```
markets, line 66: So maybe we can remove the `@current_namespace`? Not sure if this is used now.

satyakampandya: @markets `@current_namespace` is still used for actions like `Enqueue All`, `Enable All`, `Disable All`, and `Delete All`, which are applied at the namespace level.

markets: Can't we use `job.namespace` also on those actions?

satyakampandya: No, because those actions are applied to the requested namespace, which in our case can be `*`. If the namespace is `*`, it should enqueue all jobs, regardless of their specific namespace.

zedtux:
> @zedtux What do you think about these changes? Would like to hear your thoughts 👀

Thanks for asking! To me it make sense to not allow `*` as a valid namespace. 👍

markets (approved):

Issue #472, opened 2024-07-16
The `date_as_argument` option cannot be changed from `true` to `false`


iwaseasahi:
### Versions

| target | vertions |
| :--- | :--- |
| Ruby | 3.3.3 |
| Rails | 7.1.3.4 |
| sidekiq | 7.3.0 |
| sidekiq-cron | 1.12.0 |

### Summary

The `date_as_argument` option cannot be changed from `true` to `false`.

### Detail

I define the following.

`config/sidekiq_cron.yml`

```yml
send_sidekiq_metrics_job:
  cron: 0 * * * * * # Runs once per minute
  class: SendSidekiqMetricsJob
  date_as_argument: false
```

`config/initializers/sidekiq.rb`

```rb
require 'sidekiq/web'

Sidekiq.configure_server do |config|
  config.redis = {
    url: ENV.fetch('SIDEKIQ_REDIS_URL', ''),
  }

  config.on(:startup) do
    cron = YAML.load_file(File.expand_path('../../sidekiq_cron.yml', __FILE__))
    Sidekiq::Cron::Job.load_from_hash!(cron, source: 'schedule')
  end
end

Sidekiq.configure_client do |config|
  config.redis = {
    url: ENV.fetch('SIDEKIQ_REDIS_URL', ''),
  }
end
```

And when I make the following changes, the changes do not take effect.

① Define `date_as_argument` with false.（When I start sidekiq, it's `false`）

```rb
irb(main):001> Sidekiq::Cron::Job.all
2024-07-12T08:21:23.392Z pid=88304 tid=22es INFO: Sidekiq 7.3.0 connecting to Redis with options {:size=>10, :pool_name=>"internal", :url=>"redis://localhost:6379"}
=>
[#<Sidekiq::Cron::Job:0x0000000135d65e80
  ...
  @date_as_argument=false,
  @klass="SendSidekiqMetricsJob",
  ...
]
```

② `date_as_argument` be changed from false to true（When I start sidekiq, it's `true`）

```rb
irb(main):001> Sidekiq::Cron::Job.all
2024-07-12T08:22:43.534Z pid=88600 tid=22qk INFO: Sidekiq 7.3.0 connecting to Redis with options {:size=>10, :pool_name=>"internal", :url=>"redis://localhost:6379"}
=>
[#<Sidekiq::Cron::Job:0x000000012bcc9260
  ...
  @date_as_argument=true,
  @klass="SendSidekiqMetricsJob",
[...]

Issue #350, opened 2022-07-27
Broken job arguments after updating to 1.7 version (extra timestamp argument for jobs taking 0 arguments)


Azdaroth:
After updating to 1.7, we've experienced an issue with an extra argument being added for no reason - as far as the config goes, `args` is an empty array, `date_as_argument` is set to false. I suspect it's related to https://github.com/ondrejbartas/sidekiq-cron/pull/329 but haven't managed yet to identify the exact cause behinf it, especially that `enqueue_args` method returns an empty array for affected jobs.

leandro:
I can second this issue as well. Now we're getting the following error when using it:
> ArgumentError
wrong number of arguments (given 1, expected 0)

And taking one of the errors of ours as example, the array below is represents the arguments it received while it wasn't expecting any arguments:
```ruby
[1658897402.8730807]
```

nicalpi:
We have the same issue here, but reverting to 1.6.0 seems to still, somehow, send this argument now. Any one of you experience the same thing?

nicalpi:
Obviously, it was only after having made some code changes that I realised the new `date_as_argument:` in the readme. While this can be set to false, it would have been nice to have it false as a default I think to not break existing implementations.

nicalpi:
@Azdaroth @leandro if like us you upgrade to 1.7.0 before realising the new argument, then I would recommend you to delete all your existing cron jobs and restart sidekiq. This will then recreate the job now setting correctly the `      date_as_argument` as false.

I think, if like us you deployed the code before setting the `date_as_argument:` to  `false` then the jobs were now set in stone with this a lingering argument.

julianajuliano:
We had the same issue and currently have no way of updating to 1.7.0

cameronbourgeois:
It wasn't quite clear from the above comments but from what I can tell, 1.7.0 ignores date_as_argument in your config and sends the additional date parameter to the perform method anyway.

After downgrading to 1.6.0, the `@date_as_argument=false` had disappeared from my jobs, but the argument was then defined explicitly: `@args=[1660264240.1838207]`.

As @nicalpi suggested, destroying and recreating my jobs cleared this arg. This is the command I used:

```
Sidekiq::Cron::Job.destroy_all!
Sidekiq::Cron::Job.load_from_hash YAML.load_file('path/to/my_schedule.yml')
```

Once this was done, I checked my jobs and I see the argument is gone: `@args=[]`.

[...]

PR #485, opened 2024-09-15
Fix: Ensure date_as_argument option can be set from true to false in Sidekiq Cron jobs

satyakampandya:
#### Summary
This pull request fixes an issue where the `date_as_argument` option in Sidekiq Cron jobs cannot be changed from true to false once it has been set to true. The issue occurs because the Redis key for `date_as_argument` is not removed when the configuration is updated.

#### Changes
- Added logic to delete the `date_as_argument` field from Redis when `date_as_argument` is set to false.
- This ensures that stale values in Redis are cleared, allowing configuration changes to take effect as expected

Fixes: #472

### DIFF
--- a/lib/sidekiq/cron/job.rb
+++ b/lib/sidekiq/cron/job.rb
@@ -413,7 +413,7 @@ module Sidekiq
 
       # Export job data to hash.
       def to_hash
-        hash = {
+        {
           name: @name,
           namespace: @namespace,
           klass: @klass.to_s,
@@ -421,6 +421,7 @@ module Sidekiq
           description: @description,
           source: @source,
           args: @args.is_a?(String) ? @args : Sidekiq.dump_json(@args || []),
+          date_as_argument: date_as_argument? ? "1" : "0",
           message: @message.is_a?(String) ? @message : Sidekiq.dump_json(@message || {}),
           status: @status,
           active_job: @active_job ? "1" : "0",
@@ -429,18 +430,6 @@ module Sidekiq
           last_enqueue_time: serialized_last_enqueue_time,
           symbolize_args: symbolize_args? ? "1" : "0",
         }
-
-        if date_as_argument?
-          hash.merge!(date_as_argument: "1")
-        else
-          # Ensure that the `date_as_argument` field is removed from Redis when `date_as_argument` is false.
-          # This prevents stale values from persisting and ensures that changes to the config are correctly applied.
-          Sidekiq.redis do |conn|
-            conn.hdel(redis_key, 'date_as_argument')
-          end
-        end
-
-        hash
       end
 
       def errors
--- a/lib/sidekiq/cron/job.rb
+++ b/lib/sidekiq/cron/job.rb
@@ -432,6 +432,12 @@ module Sidekiq
 
         if date_as_argument?
           hash.merge!(date_as_argument: "1")
+        else
+          # Ensure that the `date_as_argument` field is removed from Redis when `date_as_argument` is false.
+          # This prevents stale values from persisting and ensures that changes to the config are correctly applied.
+          Sidekiq.redis do |conn|
+            conn.hdel(redis_key, 'date_as_argument')
+          end
         end
 
         hash
--- a/test/unit/job_test.rb
+++ b/test/unit/job_test.rb
@@ -370,7 +370,7 @@ describe "Cron Job" do
       assert_equal true, stored_job.date_as_argument?
     end
 
-    it 'does not store date_as_argument in Redis when false' do
+    it 'stores date_as_argument in Redis when false' do
       Sidekiq::Cron::Job.create(@args.merge(date_as_argument: false))
       stored_job = Sidekiq::Cron::Job.find(@args[:name])
       assert_equal false, stored_job.date_as_argument?
--- a/test/unit/job_test.rb
+++ b/test/unit/job_test.rb
@@ -354,7 +354,7 @@ describe "Cron Job" do
     end
   end
 
-  describe 'Handling date_as_argument in Sidekiq Cron Job' do
+  describe 'handling date_as_argument' do
     before do
       @args = {
         name: 'Test',
@@ -364,19 +364,19 @@ describe "Cron Job" do
       }
     end
 
-    it 'sets date_as_argument to true in the Sidekiq Cron Job' do
+    it 'sets date_as_argument to true' do
       Sidekiq::Cron::Job.create(@args.merge(date_as_argument: true))
       stored_job = Sidekiq::Cron::Job.find(@args[:name])
       assert_equal true, stored_job.date_as_argument?
     end
 
-    it 'sets date_as_argument to false in the Sidekiq Cron Job' do
+    it 'sets date_as_argument to false' do
       Sidekiq::Cron::Job.create(@args.merge(date_as_argument: false))
       stored_job = Sidekiq::Cron::Job.find(@args[:name])
       assert_equal false, stored_job.date_as_argument?
     end
 
-    it 'updates date_as_argument from true to false in the Sidekiq Cron Job' do
+    it 'updates date_as_argument from true to false' do
       Sidekiq::Cron::Job.create(@args.merge(date_as_argument: true))
       stored_job = Sidekiq::Cron::Job.find(@args[:name])
       assert_equal true, stored_job.date_as_argument?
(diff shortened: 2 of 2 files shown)


markets:
I have a question @satyakampandya about this: Why this attribute needs special treatment compared to the others https://github.com/sidekiq-cron/sidekiq-cron?tab=readme-ov-file#job-properties?

satyakampandya:
@markets If you take a look at this PR https://github.com/sidekiq-cron/sidekiq-cron/pull/392/files#diff-45ac47a9faa0840538f2cae00fdd31667a33fb6b4f94a7428d9c80bceaf8a87eL407, it introduces these special conditions to resolve a version compatibility issue. I’m not entirely sure of the reasoning behind that PR.

In this current PR, if `date_as_argument` is set to `false`, we make sure to delete the `date_as_argument` field from the `redis_key`.

markets:
Ok! I'd suggest to add a test for this change.

lib/sidekiq/cron/job.rb
```diff
@@ -432,6 +432,12 @@ def to_hash
 
         if date_as_argument?
           hash.merge!(date_as_argument: "1")
+        else
```
freemanoid, line 435: Instead of this, we could just assign a value similar to others:

E.g.
```ruby
hash = {
  ...
  date_as_argument: (date_as_argument? ? "1" : "0")
}
```

So we don't need `if date_as_argument?` condition at all.

satyakampandya: @freemanoid Please have a look at this comment https://github.com/sidekiq-cron/sidekiq-cron/pull/485#issuecomment-2351677551

satyakampandya:
I agree with @markets that, as long as PR #392 is still relevant, we need to maintain the special handling for `date_as_argument`.

However, if @iwaseasahi confirms that PR #392 is no longer applicable, we could revert to a simpler approach where `date_as_argument` is treated like other attributes,
```ruby
date_as_argument: date_as_argument? ? "1" : "0"
```

Let's await @iwaseasahi's feedback on this.

freemanoid:
> @satyakampandya @freemanoid I don't remember all the history behind #392 (it seems it fixes a quite relevant issue #350, but can't remember either all the reasoning behind).

If that special handling is only for the backward compatibility with 1.6.0 then we can ignore it in 2.0.0, right?

iwaseasahi:
@satyakampandya @markets 
Thank you for your pull request and review!
It worked correctly with this branch.

markets:
I'm a bit afraid to revert back to:

```ruby
date_as_argument: date_as_argument? ? "1" : "0"
```

As it introduced a lot of problems in the past, I think this is due people having jobs defined in this way `def perform` instead of `def perform(*args)`. That raises:

> ArgumentError
> wrong number of arguments (given 1, expected 0)

So, probably this is finally the good fix and the one we should merge!

freemanoid:
> I think this is due people having jobs defined in this way `def perform` instead of `def perform(*args)`. That raises:
> 
> > ArgumentError
> > wrong number of arguments (given 1, expected 0)
> 
> So, probably this is finally the good fix and the one we should merge!

This is unlikely, because when `date_as_argument: 0` it shouldn't pass a date to the job.

markets:
`date_as_argument: "0"` evaluates to a "truthy" value, see this comment: https://github.com/sidekiq-cron/sidekiq-cron/issues/350#issuecomment-1409798837. 

So if we revert back that code #392, we'll probably need to fix that too (~> https://github.com/sidekiq-cron/sidekiq-cron/blob/v1.6.0/lib/sidekiq/cron/job.rb#L282).

freemanoid:
> `date_as_argument: "0"` evaluates to a "truthy" value, see this comment: [#350 (comment)](https://github.com/sidekiq-cron/sidekiq-cron/issues/350#issuecomment-1409798837).

Is it true for the latest version or only for 1.6.0?

As I see this problem, in 1.6.0 we used to check only for existance of `date_as_argument`, so any value, even `"0"` was evaluated into `true`

https://github.com/sidekiq-cron/sidekiq-cron/blob/5f6bd48fbcd9cfd041394035294c3e2ecd4beab4/lib/sidekiq/cron/job.rb#L282

Then in 1.7.0 it has been fixed to properly parse only truthy values into `true`
https://github.com/sidekiq-cron/sidekiq-cron/pull/329/files#diff-45ac47a9faa0840538f2cae00fdd31667a33fb6b4f94a7428d9c80bceaf8a87eR293

And it's still how it works in current master
https://github.com/sidekiq-cron/sidekiq-cron/blob/65020e5ff3ebd849516a925206aca523682dd64c/lib/sidekiq/cron/job.rb#L56

It makes me think that the problem you're mentioning is only relevant to a migration from 1.6.0 to newer versions.

markets:
Thanks for the investigation @freemanoid! Then, it seems safe to just revert #392 🤔 

@satyakampandya Could you please take a look and update this PR accordingly? After this is done, would be nice if @iwaseasahi can make a second test with this branch.

satyakampandya:
@markets I've reverted the changes introduced in PR #392.
@iwaseasahi Could you please test this branch?

test/unit/job_test.rb
```diff
@@ -354,6 +354,39 @@
...
+      }
+    end
+
+    it 'stores date_as_argument in Redis when true' do
+      Sidekiq::Cron::Job.create(@args.merge(date_as_argument: true))
+      stored_job = Sidekiq::Cron::Job.find(@args[:name])
+      assert_equal true, stored_job.date_as_argument?
+    end
+
+    it 'does not store date_as_argument in Redis when false' do
```
freemanoid, line 373: I believe it actually stores the value

```suggestion
    it 'stores date_as_argument in Redis when false' do
```

However, if we want to check the actual value in redis, it's better to make a redis query instead of `Sidekiq::Cron::Job.find`

```ruby
hash_from_redis = Sidekiq.redis do | conn|
  conn.hgetall("cron_job:default:Test")
end

assert_equal "0", hash_from_redis["date_as_argument"]
```

iwaseasahi:
@satyakampandya @markets 
It worked correctly with this branch again.
Thanks！

markets:
> It worked correctly with this branch again.

Nice 👌 thanks for confirming @iwaseasahi!

So before merging, I think we just need to update a bit the current tests @satyakampandya.

satyakampandya:
@markets I'll update the PR tomorrow.

test/unit/job_test.rb
satyakampandya: @freemanoid @markets 

Is it a good idea to hard-code the Redis key `cron_job:default:Test`? Changes in future updates could alter the key format, causing the test to break. Instead, we could use 
```ruby
stored_job = Sidekiq::Cron::Job.find(@args[:name])
```
to avoid directly referencing the Redis key.

test/unit/job_test.rb
```diff
@@ -1460,7 +1451,7 @@
       Sidekiq.redis do |conn|
         assert_equal conn.zcard(Sidekiq::Cron::Job.new(@args).send(:job_enqueued_key)), 1, "Should have one enqueued job"
       end
-      assert_equal Sidekiq::Queue.all.first.size, 1, "Sidekiq queue 1 job in queue"
+      assert_equal Sidekiq::Queue.all.first&.size, 1, "Sidekiq queue 1 job in queue"
```
markets, line 1454: @satyakampandya Why we need this change?

satyakampandya: To address the `Method invocation 'size' may produce 'NoMethodError'` warning, this change implements safe navigation.

test/unit/job_test.rb
freemanoid: My preference is to make tests stupid and not reuse much code in tests.

test/unit/job_test.rb
markets: @satyakampandya Is this related to this PR? Not sure from where this warning is raised, but if this is unrelated to this PR, I'd like to revert those changes for now.

markets:
@satyakampandya @freemanoid I'd like to merge this one before the next release candidate. I think we all agree with the changes in `lib/sidekiq/cron/job.rb`, but I'm not sure if the current tests are ok. Some questions/suggestions:
- no need to write "Sidekiq Cron" in the titles/descriptions
- it seems we are just using the same "asserts" in all test cases 🤔

markets:
Ah! sorry didn't realize you were checking all the possible transitions.

markets (approved):

PR #487, opened 2024-09-19
Fix: Do not require "sidekiq/cron" to run "sidekiq/cron/web"

freemanoid:
Fix https://github.com/sidekiq-cron/sidekiq-cron/issues/486

### Summary

This PR solves allows running web extension without requiring sidekiq-cron functionality. This eliminates the risk of running sidekiq-cron on regular sidekiq processes unless explicitly required.

### Changes

- Decouple `sidekiq/cron/web` dependencies from `Sidekiq::Cron`
- Extract config into `Sidekiq::Cron::Config`
- Deprecate `Sidekiq::Cron.configure` in favor of `Sidekiq::Cron::Config.configure`
- Add a separate `test_helper_without_require_sidekiq_cron.rb` test helper for `web_extension_test.rb`
- Run isolated test for `web_extension_test.rb`
- Update README

### Open questions

- I don't think I need to add a mutex to `Config`, because there are only simple assignments


markets:
Hey @freemanoid 👋🏼 thank you so much for this PR, but I think there are too much changes... I thought we can do it by just changing the place of some "requires".

If can implement the same without changing all this stuff, I'm still +1, otherwise I'd prefer to not merge this PR.

Things that I don't like:
- `test/test_helper_without_require_sidekiq_cron.rb`: this is not so common, I'd prefer to keep things as simple as possible
- `Sidekiq::Cron::Config.configure`: the old api seems better and shorter to me

freemanoid:
> I thought we can do it by just changing the place of some "requires".

I'd love to do that too, but I don't see a way how it can be solved easily without decoupling Config from Sidekiq::Cron. That's the whole source of the problem and the direct solution. The reason is that before namespaces were introduced, we didn't need access to the Config object from web extension.

You're definitely more experienced than I am, so I would appreciate any guidance you can give me. 

> * `test/test_helper_without_require_sidekiq_cron.rb`: this is not so common, I'd prefer to keep things as simple as possible

This is needed because other tests require `sidekiq/cron` and thus break the idea of testing web_extension without requiring the whole gem. Since all the tests are running in the same ruby process, if one test requires sidekiq/cron, it will be loaded for all succeeding tests.

markets:
Yeah, sorry I'm also unsure how this can be simplified, but I'm not interested in working on this at this moment, probably the use-case is too much specific? I mean, requiring the library to be able to use the UI part seems reasonable. 

As I said before, if we can decouple it with very small changes, I'd be ok to merge it too, otherwise, I'd prefer to keep as simple as it's now.

freemanoid:
> probably the use-case is too much specific

My use case is that I don't want to run sidekiq-cron on every sidekiq process. Thus I have several types of instances:
- regular sidekiq process: has no `require "sidekiq/cron"`
- small sidekiq-cron instance: has `require "sidekiq/cron"`
- web server which also serves sidekiq web UI: has no`require "sidekiq/cron"`, only `require "sidekiq/cron/web"` in `routes.rb`

If I'd have to add `require "sidekiq/cron"` to `routes.rb` to make the web ui work, it will be automatically loaded for the regular sidekiq processes and turn any sidekiq process into sidekiq-cron, which is inefficient given that I have thousands of sidekiq processes.

markets:
If you want to avoid all the Redis load that comes from Sidekiq-Cron, you can also use `Sidekiq::Options[:cron_poll_interval] = 0` in those processes (see more in this section https://github.com/sidekiq-cron/sidekiq-cron?tab=readme-ov-file#under-the-hood):

> When running with many Sidekiq processes, the polling can add significant load to Redis. You can disable polling on some processes by setting Sidekiq::Options[:cron_poll_interval] = 0 on these processes.

freemanoid:
> If you want to avoid all the Redis load that comes from Sidekiq-Cron, you can also use `Sidekiq::Options[:cron_poll_interval] = 0`

Thank you. That sounds like a compromise that is good enough for me. Then we need to update the readme to indicate that and probably add this information to the changelog or 2.0 migration notes.

markets:
You're welcome @freemanoid! In fact, this is already in the README, since 2 years ago 👉🏼 #368.

Feel free to close this PR.

markets:
Ah yes! That's right! Feel free to send a patch, otherwise I'll try to add it by myself before releasing.

PR #492, opened 2024-10-02
Fix redirection issues in web extension for namespace-specific routes

satyakampandya:
This PR addresses a redirection issue in the Sidekiq Cron web extension where redirects were not consistently pointing to namespace-specific routes. The changes ensure that after performing operations (`enqueue`, `delete`, `enable`, `disable`) on cron jobs, the user is redirected to the appropriate namespace view.

### DIFF
--- a/lib/sidekiq/cron/web_extension.rb
+++ b/lib/sidekiq/cron/web_extension.rb
@@ -56,24 +56,24 @@ module Sidekiq
 
         # Enqueue cron job.
         app.post '/cron/namespaces/:namespace/jobs/:name/enque' do
-          if job = Sidekiq::Cron::Job.find(route_params[:name], route_params[:namespace])
+          if (job = Sidekiq::Cron::Job.find(route_params[:name], route_params[:namespace]))
             job.enque!
           end
-          redirect params['redirect'] || "#{root_path}cron"
+          redirect params['redirect'] || "#{root_path}cron/namespaces/#{route_params[:namespace]}"
         end
 
         # Delete all schedules.
         app.post '/cron/namespaces/:namespace/all/delete' do
           Sidekiq::Cron::Job.all(route_params[:namespace]).each(&:destroy)
-          redirect "#{root_path}cron/namespaces/#{route_params[:namespace]}"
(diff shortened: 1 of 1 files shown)


lib/sidekiq/cron/web_extension.rb
```diff
@@ -56,24 +56,24 @@ def self.registered(app)
 
         # Enqueue cron job.
         app.post '/cron/namespaces/:namespace/jobs/:name/enque' do
-          if job = Sidekiq::Cron::Job.find(route_params[:name], route_params[:namespace])
+          if (job = Sidekiq::Cron::Job.find(route_params[:name], route_params[:namespace]))
```
markets, line 59: Is this change necessary? I'd prefer to avoid unnecessary diff 🙏

lib/sidekiq/cron/web_extension.rb
```diff
@@ -56,24 +56,24 @@ def self.registered(app)
...
             job.enque!
           end
-          redirect params['redirect'] || "#{root_path}cron"
+          redirect params['redirect'] || "#{root_path}cron/namespaces/#{route_params[:namespace]}"
```
markets, line 62: Maybe we can create a helper to DRY this code? Something like "redirect_with_namespace".

lib/sidekiq/cron/web_extension.rb
satyakampandya: Not necessary, just following best practice guidelines, to add clarifying parentheses.

lib/sidekiq/cron/web_extension.rb
satyakampandya: Should we implement the `redirect_with_namespace` method inline within the `WebExtension `like this?
```ruby
app.helpers do
  def redirect_with_namespace
    "#{root_path}cron/namespaces/#{route_params[:namespace]}"
  end
end
```
Or would it be better to create a separate helper file?

markets: I think it's fine for now keep it inline. We can create a separate module in the future when more helpers are implemented.

markets (approved):

markets:
Thanks @satyakampandya 🙌

PR #495, opened 2024-10-06
Refactor sidekiq cron configurations

satyakampandya:
This PR refactors the configuration management in `sidekiq-cron` by moving `cron_poll_interval` and other related options from `Sidekiq::Options` to the `Sidekiq::Cron.configuration`. This change improves separation of concerns by allowing `sidekiq-cron` to manage its own settings independently from sidekiq's core options.

### DIFF
--- a/lib/sidekiq-cron.rb
+++ b/lib/sidekiq-cron.rb
@@ -1,2 +1,7 @@
 require "sidekiq"
 require "sidekiq/cron"
+require "sidekiq/cron/job"
+require "sidekiq/cron/namespace"
+require "sidekiq/cron/poller"
+require "sidekiq/cron/launcher"
+require "sidekiq/cron/schedule_loader"
--- a/lib/sidekiq/cron.rb
+++ b/lib/sidekiq/cron.rb
@@ -9,6 +9,10 @@ module Sidekiq
       yield(configuration) if block_given?
     end
 
+    def self.reset!
+      self.configuration = Configuration.new
+    end
+
     class Configuration
       # The default namespace is used when no namespace is specified.
       attr_accessor :default_namespace
--- a/lib/sidekiq/cron.rb
+++ b/lib/sidekiq/cron.rb
@@ -36,6 +36,10 @@ module Sidekiq
       # This value controls how many past job executions are stored.
       attr_accessor :cron_history_size
 
+      # The interval, in seconds, at which to poll for scheduled cron jobs.
+      # This determines how frequently the scheduler checks for jobs to enqueue.
+      attr_accessor :cron_poll_interval
+
       # The path to a YAML file containing multiple cron job schedules.
       # This file should use the same format as Resque-scheduler for job definitions.
       attr_accessor :cron_schedule_file
@@ -45,6 +49,7 @@ module Sidekiq
         @natural_cron_parsing_mode = :single
         @reschedule_grace_period = 60
         @cron_history_size = 10
+        @cron_poll_interval = 30
         @cron_schedule_file = 'config/schedule.yml'
       end
 
--- a/lib/sidekiq/cron.rb
+++ b/lib/sidekiq/cron.rb
@@ -33,10 +33,15 @@ module Sidekiq
       # jobs that missed their schedules during the deployment. E.g., jobs that run once a day.
       attr_accessor :reschedule_grace_period
 
+      # The maximum number of recent cron job execution histories to retain.
+      # This value controls how many past job executions are stored.
+      attr_accessor :cron_history_size
+
       def initialize
         @default_namespace = 'default'
         @natural_cron_parsing_mode = :single
         @reschedule_grace_period = 60
+        @cron_history_size = 10
       end
 
       def natural_cron_parsing_mode=(mode)
--- a/lib/sidekiq/cron.rb
+++ b/lib/sidekiq/cron.rb
@@ -2,7 +2,6 @@ require "sidekiq/cron/job"
 require "sidekiq/cron/namespace"
 require "sidekiq/cron/poller"
 require "sidekiq/cron/launcher"
-require "sidekiq/cron/schedule_loader"
 
 module Sidekiq
   module Cron
@@ -37,11 +36,16 @@ module Sidekiq
       # This value controls how many past job executions are stored.
       attr_accessor :cron_history_size
 
+      # The path to a YAML file containing multiple cron job schedules.
+      # This file should use the same format as Resque-scheduler for job definitions.
+      attr_accessor :cron_schedule_file
+
       def initialize
         @default_namespace = 'default'
         @natural_cron_parsing_mode = :single
         @reschedule_grace_period = 60
         @cron_history_size = 10
+        @cron_schedule_file = 'config/schedule.yml'
       end
 
       def natural_cron_parsing_mode=(mode)
--- a/lib/sidekiq/cron.rb
+++ b/lib/sidekiq/cron.rb
@@ -1,8 +1,3 @@
-require "sidekiq/cron/job"
-require "sidekiq/cron/namespace"
-require "sidekiq/cron/poller"
-require "sidekiq/cron/launcher"
-
 module Sidekiq
   module Cron
     class << self
--- a/lib/sidekiq/cron/launcher.rb
+++ b/lib/sidekiq/cron/launcher.rb
@@ -8,14 +8,13 @@ require 'sidekiq/cron/poller'
 module Sidekiq
   module Cron
     module Launcher
-      DEFAULT_POLL_INTERVAL = 30
 
       # Add cron poller to launcher.
       attr_reader :cron_poller
 
       # Add cron poller and execute normal initialize of Sidekiq launcher.
       def initialize(config, **kwargs)
-        config[:cron_poll_interval] = DEFAULT_POLL_INTERVAL if config[:cron_poll_interval].nil?
+        config[:cron_poll_interval] = Sidekiq::Cron.configuration.cron_poll_interval.to_i
 
         @cron_poller = Sidekiq::Cron::Poller.new(config) if config[:cron_poll_interval] > 0
         super
--- a/lib/sidekiq/cron/launcher.rb
+++ b/lib/sidekiq/cron/launcher.rb
@@ -1,4 +1,3 @@
-require 'sidekiq/cron'
 require 'sidekiq/cron/poller'
 
 # For Cron we need to add some methods to Launcher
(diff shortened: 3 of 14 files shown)


markets:
Nice job @satyakampandya 🙌

I think we should update the README too. In some points we still mention the old config approach.

test/unit/launcher_test.rb
```diff
@@ -3,12 +3,12 @@
 describe 'Cron launcher' do
   describe 'initialization' do
     before do
-      Sidekiq::Options[:cron_poll_interval] = nil
+      Sidekiq::Cron.configuration.cron_poll_interval = 30
```
markets, line 6: Maybe we can remove this block? That's already the default value.

lib/sidekiq/cron.rb
```diff
@@ -2,7 +2,6 @@
 require "sidekiq/cron/namespace"
 require "sidekiq/cron/poller"
 require "sidekiq/cron/launcher"
-require "sidekiq/cron/schedule_loader"
```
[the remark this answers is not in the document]
satyakampandya: I removed it to avoid a cyclic require issue. It'll be great if you could share best solution for it. 
If I add it again, I'm facing `undefined method `configuration' for Sidekiq::Cron:Module`.

test/unit/launcher_test.rb
satyakampandya: It fixes some failing tests on CI.

lib/sidekiq/cron.rb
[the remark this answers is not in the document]
markets: Ok, I understand, but we'll need to find a solution. Otherwise the automatic scheduler loader is not going to work anymore. Maybe we can move the `Configuration` class outside the main module 🤔?

test/unit/launcher_test.rb
markets: But if we are checking the defaults here, I think we should avoid to explicitly set 30, but instead run `Sidekiq::Cron.configure`?

lib/sidekiq/cron.rb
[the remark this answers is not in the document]
satyakampandya: Yes, I'll have a look tonight. If you've any suggestions please let me know.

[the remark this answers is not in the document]
markets: The point is that we need to run the default config, before loading other things. So another idea is to move these `require`s at the end of the file, after running the `Sidekiq::Cron.configure` (which will create the default config).

test/unit/launcher_test.rb
satyakampandya: Yes, I tried it, locally it works, but it fails here on CI tests.

lib/sidekiq/cron.rb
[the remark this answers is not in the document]
markets: Yeah, that's a quick idea that came out of my head.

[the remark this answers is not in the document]
satyakampandya: It'll still cause cyclic require issue, right?

satyakampandya:
@markets, I'm having some trouble with the test cases — everything seems to work fine on my local setup. Could you help me figure out what's going wrong? I'd really appreciate it!

markets:
Ok! I'll take a look in the following days. Could you please update your branch with current master 🙏🏼?

markets:
@satyakampandya Just took a quick look using your branch locally and I can reproduce it intermittently. If I force the tests order (with this patch 👇🏼) it always passes:

```ruby
class Minitest::Test
  def self.test_order
   :alpha
  end
end
```

So clearly this depends on the order of execution. Still not sure how to fix it, but we probably need to re-initialize the default config on each test (via a `before` block).

satyakampandya:
Sure @markets, I'll have a look at it.

satyakampandya:
@markets Do you think it's due to below line of code, which doesn't create new Configuration if it's already there?
```ruby
def self.configure
  self.configuration ||= Configuration.new
  yield(configuration) if block_given?
end
```

markets:
hmm 🤔 I think it's more due to the execution order. By default, tests run in random order and the issue can be reproduced locally. Forcing the order (as described above) to be static, makes the suite to pass ✅. So my bet: in other test cases we are changing some settings and we don't restore them. We should run each test case within an expected configuration (the defaults looks like the best idea).

satyakampandya:
Even when we set `Sidekiq::Cron.configure` in each test, the `initialize` method of `Sidekiq::Cron` is only called once because of the line `self.configuration ||= Configuration.new`. This means the configuration is not reset between tests as `||=` only assigns a new Configuration if it's nil.

markets:
In that case, we can create a `reset!` method to return back to defaults.

satyakampandya:
@markets Are you suggesting it like this?
[refactor-sidekiq-cron-configuration.patch](https://github.com/user-attachments/files/17344697/refactor-sidekiq-cron-configuration.patch)

markets:
Yes! but the `reset!` method doesn't need to yield anything, just return back to defaults.

lib/sidekiq-cron.rb
```diff
@@ -1,2 +1,7 @@
 require "sidekiq"
 require "sidekiq/cron"
+require "sidekiq/cron/job"
+require "sidekiq/cron/namespace"
+require "sidekiq/cron/poller"
+require "sidekiq/cron/launcher"
+require "sidekiq/cron/schedule_loader"
```
satyakampandya, line 7: @markets Please verify this change.

markets: What do you mean by verify 🤔? Could you please verify it by yourself in your apps too?

satyakampandya: It works fine with my app.

lib/sidekiq/cron/launcher.rb
```diff
@@ -1,3 +1,4 @@
+require 'sidekiq/cron'
```
markets, line 1: Is this require necessary? I think we can ✂️

test/unit/web_extension_test.rb
```diff
@@ -12,6 +12,7 @@ def app
   before do
     env 'rack.session', { csrf: TOKEN }
     env 'HTTP_X_CSRF_TOKEN', TOKEN
+    Sidekiq::Cron.reset!
```
markets, line 15: Maybe we can remove it from here? 🤔 I'm almost sure in this test we don't modify the settings.

lib/sidekiq/cron/launcher.rb
satyakampandya: Won't we need it to use `Sidekiq::Cron.configuration.cron_poll_interval.to_i` in `initialize`?

test/unit/web_extension_test.rb
satyakampandya: Sure

lib/sidekiq/cron/launcher.rb
markets: But in the entry point `lib/sidekiq-cron.rb` of the library we already load all the modules:

```ruby
require "sidekiq"
require "sidekiq/cron"
require "sidekiq/cron/job"
require "sidekiq/cron/namespace"
require "sidekiq/cron/poller"
require "sidekiq/cron/launcher"
require "sidekiq/cron/schedule_loader"
```

So when this file is required, the `sidekiq/cron` was loaded too, right?

satyakampandya: In that case, how about removing below require from `lib/sidekiq/cron/job.rb`?
```
require 'sidekiq'
require 'sidekiq/cron'
require 'sidekiq/cron/support'
```

markets:
Sure @satyakampandya! No rush! This branch is improving a lot our codebase 👌

This is last big change I'd like to have for the final v2. And my plan is to release a rc2 with that, and then,  the v2 before the end of the year.

satyakampandya:
Hey @markets,

We need to find a way to reload schedules from `cron_schedule_file`. Currently, when the gem is loaded, it pulls the default file from `config/sidekiq.yml`. However, if we update it dynamically using the following code in an initializer:

```ruby
Rails.application.reloader.to_prepare do
    config.cron_schedule_file = 'config/schedule1.yml'
  end
end
```
The schedule loader doesn’t trigger automatically. We need to explore a solution for reloading the schedule when the file is updated.

I think we should consider removing the auto-loading behavior below, or find a way to prevent loading the default file if the user is updating it dynamically. The current setup overrides existing loaded jobs, which can cause issues. We need to ensure that if a custom `cron_schedule_file` is set, it doesn't get overridden by the default.

```ruby
Sidekiq.configure_server do |config|
  schedule_file = Sidekiq::Cron.configuration.cron_schedule_file

  if schedule_file
    config.on(:startup) do
      Sidekiq::Cron::ScheduleLoader.load!(schedule_file)
    end
  end
end
```

Check changes outlined in attached patch. 
[schedule-file.patch](https://github.com/user-attachments/files/17350234/schedule-file.patch)

Let me know your thoughts!

markets:
Is that something introduced in this branch? If not, please 🙏 open a new issue, so we can focus this PR on the configuration refactor.

satyakampandya:
Yes, it is part of this PR.

When we moved `cron_schedule_file` from `Sidekiq::Options` to `Sidekiq::Cron.configure`, the custom cron_schedule_file stopped loading automatically. I can reproduce this behavior in the master branch as well. It seems that this issue has always been there—simply specifying a custom file path wasn't enough, and it still needed to be loaded manually to support custom schedule path.

We can either skip moving `cron_schedule_file` in this PR OR address the issue by implementing a fix to ensure the custom path gets loaded correctly. Let me know how you'd like to proceed!

markets:
I'm not sure if this is a bug or a feature.

But anyway, if that is something that already happens in `master`, I think it will be better to open a new discussion, as this PR is big enough.

satyakampandya:
Hey @markets,

I’ve removed the documentation related to `cron_schedule_file` configuration because it isn’t working as expected. The automatic loading feature for custom schedule files isn’t functioning properly, so it's best to exclude it from the docs for now to avoid confusion.

satyakampandya:
I can revert the changes from the last commit regarding the README if needed. I'm not entirely sure what you're suggesting—could you please clarify?

markets:
Yes please! This is out of the scope of this PR, which is big enough and with more than 40 comments...

The current behaviour is probably the expected one (see for example: https://github.com/sidekiq-cron/sidekiq-cron/pull/337#issuecomment-1844201808). You already pointed out that it works this way on `master`. Here we are only refactoring the conf module, so please please let's focus on that and avoid starting new discussions.

markets:
Hello @satyakampandya I made a first look at the whole PR and it seems good now ✅. I'll perform a full review in the following days.

Thanks!

markets (approved):

PR #506, opened 2024-10-23
Review web UI translations for multiple locales

satyakampandya:


markets:
Looks good @satyakampandya, but I think we should keep the previous translations made by humans. For example, in `id` and using Google translate:

- BEFORE => 'Last enqueued': Terakhir kali ditambahkan ke antrian ~> "Last time added to queue"
- AFTER => LastEnqueued: Terakhir Diantri ~> "Last in line"

As you can see, that's not exactly the same... I'd like to keep previous values in existing keys. So, in this case it should be:

```
LastEnqueued: Terakhir kali ditambahkan ke antrian
```

markets (approved):
Thanks @satyakampandya 🙌

PR #515, opened 2024-11-12
fix: Add quotes to Japanese translations containing multi-byte symbols

iwaseasahi:
## Description

- Enclose Japanese translation strings with double quotes to prevent YAML parse errors
- Double quotes are necessary when strings contain multi-byte symbols like `？`
- Fixes the "found character that cannot start any token" error

Resolves https://github.com/sidekiq-cron/sidekiq-cron/issues/514

## Screenshot

スクリーンショット 2024-11-12 16 20 25

### DIFF
--- a/lib/sidekiq/cron/locales/ja.yml
+++ b/lib/sidekiq/cron/locales/ja.yml
@@ -1,8 +1,8 @@
 ja:
-  AreYouSureDeleteCronJob: 本当に%{job}のcronジョブを削除しますか？
-  AreYouSureDeleteCronJobs: 本当にすべてのcronジョブを削除しますか？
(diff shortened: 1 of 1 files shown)


ise:
> Double quotes are necessary when strings contain multi-byte symbols like ？

The PR itself seems fine, but I think the cause of the error is the leading `%`, not the `？`.

```rb
> YAML.load("hoge: aaa？")
=> {"hoge"=>"aaa？"}
> YAML.load("hoge: %aaa")
(venison):86:in `<main>': (<unknown>): found character that cannot start any token while scanning for the next token at line 1 column 7 (Psych::SyntaxError)
> YAML.load("hoge: a%aa")
=> {"hoge"=>"a%aa"}
```

PR #530, opened 2025-03-06
Fix CI

leoarnold:
This fixes CI for Sidekiq 7 where otherwise `Sidekiq::ActiveJob::Wrapper` would not be defined.


markets:
I think we can close this one right @leoarnold. It seems superseded by #531.

PR #531, opened 2025-03-06
Add support for Sidekiq v8

### LINKED ISSUE #529: Sidekiq v8.0: undefined method 'settings' for class Sidekiq::Web::Application
I have updated Sidekiq to v8.0.0, and I'm using sidekiq-cron 2.1.0. I get this issue:

`
undefined method 'settings' for class Sidekiq::Web::Application
`

on the line "require 'sidekiq/cron/web'"

```ruby
require 'sidekiq/web'
require 'sidekiq/cron/web'
mount Sidekiq::Web => '/sidekiq'
````

### PR DESCRIPTION
**Pull request by leoarnold**
This change will make `sidekiq-cron` functionally compatible with Sidekiq v8.0.0.

Some users may take issue to the placment of the buttons on the web UI
due to `.pull-right` not working anymore. This is left to another developer
with better knowledge of web design.

Screenshot_2025-03-06_22-22-11

Fixes #529

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: ["2.7", "3.0", "3.1", "3.2", "3.3", "3.4", "ruby-head"]
+        ruby: ["3.2", "3.3", "3.4", "ruby-head"]
         sidekiq: ["~> 6", "~> 7"]
     services:
       redis:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,20 @@ jobs:
       fail-fast: false
       matrix:
         ruby: ["3.2", "3.3", "3.4", "ruby-head"]
-        sidekiq: ["~> 6", "~> 7"]
+        sidekiq: ["~> 6", "~> 7", "~> 8"]
+        include:
+          - ruby: "2.7"
+            sidekiq: "~> 6"
+          - ruby: "2.7"
+            sidekiq: "~> 7"
+          - ruby: "3.0"
+            sidekiq: "~> 6"
+          - ruby: "3.0"
+            sidekiq: "~> 7"
+          - ruby: "3.1"
+            sidekiq: "~> 6"
+          - ruby: "3.1"
+            sidekiq: "~> 7"
     services:
       redis:
         image: redis
--- a/Gemfile
+++ b/Gemfile
@@ -6,4 +6,4 @@ gemspec
 gem "sidekiq", ENV.fetch("SIDEKIQ_VERSION", ">= 6")
 
 # To test different Active Job versions
-gem "activejob", ENV.fetch("ACTIVE_JOB_VERSION", "~> 7")
+gem "rails", ENV.fetch("ACTIVE_JOB_VERSION", "~> 7")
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

File: .github/workflows/ci.yml [hunk 1]
```diff
@@ -12,8 +12,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: ["2.7", "3.0", "3.1", "3.2", "3.3", "3.4", "ruby-head"]
-        sidekiq: ["~> 6", "~> 7"]
+        ruby: ["3.2", "3.3", "3.4", "ruby-head"]
```
markets, line 15: I'd like to keep compatibility with those Rubies, for people still on Sidekiq v6 or v7 it makes sense.

Gemfile
```diff
@@ -3,7 +3,7 @@ source 'https://rubygems.org'
 gemspec
 
 # To test different Sidekiq versions
-gem "sidekiq", ENV.fetch("SIDEKIQ_VERSION", ">= 6")
+gem "sidekiq", ENV.fetch("SIDEKIQ_VERSION", "~> 6")
```
markets, line 6: Is this change really necessary?

Gemfile
```diff
@@ -3,7 +3,7 @@ source 'https://rubygems.org'
...
 
 # To test different Active Job versions
-gem "activejob", ENV.fetch("ACTIVE_JOB_VERSION", "~> 7")
+gem "rails", ENV.fetch("ACTIVE_JOB_VERSION", "~> 7")
```
markets, line 9: If we change this to rails, the comment and the ENV var should be updated too.

Gemfile
leoarnold: Nope, I mixed something up here.

markets:
Looks quite good now @leoarnold 👌🏼

Could you please add a screenshot to see how it looks in Sidekiq v8?

test/test_helper.rb
```diff
@@ -25,6 +25,8 @@
 require './test/support/classes'
 require './test/support/helpers'
 
+require "rails/engine/railties"
+require "sidekiq/rails"
```
markets, line 29: Could you please move these 2 `require`s some lines above? It seems more consistent to load external deps in the same place.

leoarnold:
@markets Done

lib/sidekiq/cron/web.rb
```diff
@@ -3,7 +3,16 @@
 require "sidekiq/cron/namespace"
 
 if defined?(Sidekiq::Web)
-  if Gem::Version.new(Sidekiq::VERSION) >= Gem::Version.new('7.3.0')
+  if Gem::Version.new(Sidekiq::VERSION) >= Gem::Version.new('8.0.0')
```
georgebancila, line 6: maybe add a helper method where you send the wanted version?

```
def sidekiq_version_check(version)
    Gem::Version.new(Sidekiq::VERSION) >= Gem::Version.new(version)
end
```

georgebancila (approved):

lib/sidekiq/cron/web.rb
markets: We have different checks like this one around the code base, but in different type of files: views, Ruby files, ... Not sure if we have a place to add a helper available in all those scenarios.

markets (approved):
@leoarnold Thanks! I think we can merge this one and work on refining the UI in new branches.

PR #534, opened 2025-03-08
Add fallback to `.yaml` extension

### LINKED ISSUE #528: `.yaml` extension not loading
I just spent few minutes confused why the jobs are not scheduled and found out that the default YAML file name is `schedule.yml` and `schedule.yaml` is not loading. I understand it is just a default value, but it would be would be nice if `.yaml` also works, as it is [the official YAML extension](https://yaml.org/faq.html).

krystof-k:
Try loading the file with the `.yaml` extension if the original file doesn't exist. Closes #528.


markets:
@krystof-k Could you please update from master? We recently merged Sidekiq v8 support.

krystof-k:
@markets done!

markets (approved):

PR #535, opened 2025-03-08
Always show parsed cron expression in web UI

### LINKED ISSUE #532: Show parsed cron expression in web UI when used natural language
Currently if natural language definition is used, it is rendered twice in the web UI:

Image

It would be better to render the parsed cron expression as the secondary `<small>` text in that case. I think it is this line:

https://github.com/sidekiq-cron/sidekiq-cron/blob/6fff355eb8701ec910d831412286c4244d81c5c9/lib/sidekiq/cron/views/cron.erb#L79

krystof-k:
Always show natural-language cron and parsed cron expression below in the web UI. Fixes #532.


markets:
@krystof-k Could you please update from master? We recently merged Sidekiq v8 support.

krystof-k:
@markets done!

markets (approved):

PR #538, opened 2025-03-09
Properly support Sidekiq v8 new UI

### LINKED ISSUE #537: Improve UI on Sidekiq v8
We recently merged Sidekiq v8 support (https://github.com/sidekiq-cron/sidekiq-cron/pull/531), but the UI needs some love.

eqladios:
Sidekiq v8 dropped Bootstrap from its assets, so I tried to fix the UI using the same classes used in sidekiq v8 UI, I opted to creating a new v8 folder in views, changes are too many to put conditional rendering in the current views

A more elegant solution would be defining a CSS asset for sidekiq-cron and load it when registering the extension in sidekiq.

- use separate folder for views for >= v8
- use vanilla CSS classes defined by sidekiq UI and drop Bootstrap CSS classes

# Screenshots
image
image
image

fixes #537

### DIFF
--- a/lib/sidekiq/cron/views/cron.erb
+++ b/lib/sidekiq/cron/views/cron.erb
@@ -1,34 +1,31 @@
-<header class='row'>
-  <div class='col-sm-5 pull-left'>
-    <h3>
+<section>
+  <header>
+    <h2>
       <%= t('CronJobs') %>
-      <small><%= @current_namespace %></small>
-    </h3>
-  </div>
-  <div class='col-sm-7 pull-right h2'>
+      <small>(<%= @current_namespace %>)</small>
+    </h2>
     <% if @cron_jobs.size > 0 %>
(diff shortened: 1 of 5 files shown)


markets:
Hey @eqladios thanks a lot for working on this! 

Maybe we can invert the logic and make the v8 templates the default ones? In future major versions (still not planned) I'd like to remove support for <v8.

eqladios:
Hi @markets, I created a folder named `legacy` for the current templates and left the new ones under `/views`

I also changed the buttons in the `<td>` in the cron table as `btn-primary` was showing weird on hover, so I used `btn-warn` instead as `btn-warn` and `btn-danger` in `td` are handled correctly on hover in sidekiq css; as I said a more elegant solution would be for sidekiq-cron to define its own css

image

image

image

markets (approved):

leoarnold:
Very nice :heart_eyes:

PR #541, opened 2025-03-12
Refactor constantize helper to use `Object.const_get`

dmeremyanin:
This update simplifies the constantize helper method by utilizing `Object.const_get`. Similar changes have been made in Rails and Sidekiq, as shown in the following PR and commit:
* https://github.com/rails/rails/pull/41877
* https://github.com/sidekiq/sidekiq/commit/68f7b7fffb13162328a8bee60106065911346c99

Since `sidekiq-cron` requires Ruby 2.7 or later, we can adopt this approach as well.


markets (approved):
Nice! Thanks @dmeremyanin 🙌🏼

PR #543, opened 2025-03-15
Allow testing of schedule by library consumers

tttffff:
### Motivation / Background

After accidently adding a typo in a class name on the schedule, and [thankfully] having it caught in a PR, I wanted a better way to test the schedule in my project.

This Pull Request has been created because it enables the same schedule loader that runs in production to be called in a testing environment. This way, we have the ability to show that there are no configuration errors.

### Detail

This Pull Request changes the implementation of ScheduleLoader to enable consumers of the library to test their schedule in an environment where redis is running but the Sidekiq service is not (this is usual for testing environments).

As this change is purely an implementation detail, I have not added or changed any tests.

### Additional information

I have added an example of how to test projects in the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -511,6 +511,45 @@ Sidekiq::Cron.configure do |config|
 end
 ```
 
+## Testing your configuration
+
+You can test your application's configuration by loading the schedule in your test suite. Below is an example using RSpec in a Rails project:
+
+```ruby
+# spec/cron_schedule_spec.rb
+require "rails_helper"
+
+RSpec.describe "Cron Schedule" do
+  let(:schedule_loader) { Sidekiq::Cron::ScheduleLoader.new }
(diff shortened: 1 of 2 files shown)


markets (approved):
Looks good 👌 Thanks!

cswilliams:
I tried out the sample specs from the `README.md` on the latest version of sidekiq-cron and I get this error:
```ruby
1) config/schedule.yml does not return any errors
     Failure/Error: expect(schedule_loader.load).to be_empty

     NoMethodError:
       private method 'load' called for an instance of Sidekiq::Cron::ScheduleLoader
     # ./spec/jobs/cron/schedule_yml_spec.rb:16:in 'block (2 levels) in <top (required)>'
...
```

Should it be `schedule_loader.load_schedule` instead?

tttffff:
@cswilliams That's correct. Thank you.

PR #550, opened 2025-03-22
Load schedule file within sidekiq callback

### LINKED ISSUE #507: Sidekiq::Cron.configure cron_schedule_file does not work as described in the readme for version 2.0.0.rc2
_(I'm going to post a bit of extraneous info because it might help someone else that is confused a bit by the readme and might make the same assumptions I did. I'm also not sure if this issue applies to any of the other config variables as well, but defaults work for me so I haven't tested beyond cron_schedule_file yet)_

I initially tried setting up my `config/initializers/sidekiq-cron.rb` file like so:

```ruby
  if Rails.env.production?
    Rails.logger.info "\n========= Sidekiq::Cron - loading production schedule"
    config.cron_poll_interval = 60 # Default is 30
    config.cron_schedule_file = 'config/schedules/jobs_production.yml'
  elsif Rails.env.staging?
    Rails.logger.info "\n========= Sidekiq::Cron - loading staging schedule"
    config.cron_schedule_file = 'config/schedules/jobs_staging.yml' 
  elsif Rails.env.development?
    Rails.logger.info "\n========= Sidekiq::Cron - loading development schedule"
    config.cron_schedule_file = 'config/schedules/jobs_development.yml'
  end
end
```

But this would never load the appropriate cron jobs defined in the environment specific yml files. No errors or other messages indicating failure.

On a whim I renamed one of the yml files to the default "config/schedule.yml" and restarted sidekiq, and it worked (created the defined cron job and began executing it at defined interval).

Then I thought mabye it was the subdirectory, so I tried using:
`config.cron_schedule_file = 'config/jobs_schedule_development.yml'`

And this does NOT work either. 

So with 2.0.0.rc2 (which if you want to follow the readme, you have to use instead of 1.12 to do the the Sidekiq::Cron.configure initializer at all, per [this issue](https://github.com/sidekiq-cron/sidekiq-cron/issues/469) ) the only schedule file that works is the default `config/schedule.yml`.

Side note/gotcha: cron jobs scheduled with `config/schedule.yml` (or maybe with any use of `Sidekiq::Cron` ???) do not get deleted when sidekiq restarts, even if you delete the schedule file. You have to explicitly destroy the job somewhere in your code.
Note also that `Sidekiq::Cron::Job.destroy_all!`  destroys ALL jobs, not just those scheduled by sidekiq-cron, so if you have a pending perform_later on a non-cron job somewhere before you call that (e.g. in other startup config/intiializers), you will destroy that explicit job perform_later call too.
____________________________

Ultimately the solution that worked for me was to piggyback in my `config/initializers/sidekiq.rb` and manually load the schedule yml file I needed, like so:
```ruby
Sidekiq.configure_server do |config|
  if Rails.env.production?
    schedule_file = "config/schedules/jobs_production.yml"
  elsif Rails.env.staging?
    schedule_file = "config/schedules/jobs_staging.yml"
  elsif Rails.env.development?
    schedule_file = "config/schedules/jobs_development.yml"
  else
    schedule_file = "config/schedules/jobs_test.yml"
  end

  config.on(:startup) do
    # Destroy any previously scheduled jobs
    Sidekiq::Cron::Job.destroy_all! # WARNING: This apparently destroys ALL pending jobs, not just those scheduled by sidekiq-cron!!

    if File.exist?(schedule_file)
      Rails.logger.info "\n========= Sidekiq - loading jobs schedule from #{schedule_file}"
      schedule = YAML.load_file(schedule_file)
      if schedule.present?
        Sidekiq::Cron::Job.load_from_hash!(schedule, source: "schedule")
      else
        Rails.logger.warn "\n========= Sidekiq - No job definitions found in #{schedule_file}! Skipping cron scheduling."
      end
    end

    # Any other non-cron startup jobs you need
    OtherStartupJob.perform_later
  end
```

Using this setup, it seems to reliably handle restarts in a way you'd expect if you make changes to the scheduling files.

tttffff:
Fixes #507

I believe this is the correct implementation. It moves all schedule loading to happen on sidekiq startup which is after the initializers have ran. This means that any change to `config.cron_schedule_file` in an initializer will be respected.

**However, please consider the very real possibility that someone has created a `config/schedule.yml` file but has then blindly copied the configuration from [here](https://github.com/sidekiq-cron/sidekiq-cron/?tab=readme-ov-file#configuration). They may not not have a `config/my_schedule.yml` file but this does not currently matter as that setting is effectively ignored.**

In the above example, that person may upgrade their sidekiq-cron gem only to find that their schedule stops being loaded. As such, I think there is a decision to be made. Should we follow this implementation that is correct, or write more logic to protect anyone who has done the above? I personally think we should move forward with this approach and explain in the changelog that configs should be checked.

### DIFF
--- a/lib/sidekiq/cron/schedule_loader.rb
+++ b/lib/sidekiq/cron/schedule_loader.rb
@@ -49,8 +49,9 @@ module Sidekiq
 end
 
 Sidekiq.configure_server do |config|
-  schedule_loader = Sidekiq::Cron::ScheduleLoader.new
-  next unless schedule_loader.has_schedule_file?
-
-  config.on(:startup) { schedule_loader.load_schedule }
+  config.on(:startup) do
+    schedule_loader = Sidekiq::Cron::ScheduleLoader.new
+    next unless schedule_loader.has_schedule_file?
+    schedule_loader.load_schedule
+  end
 end
--- a/test/unit/schedule_loader_test.rb
+++ b/test/unit/schedule_loader_test.rb
@@ -4,16 +4,19 @@ describe 'ScheduleLoader' do
   before do
     Sidekiq::Cron.reset!
     Sidekiq::Options[:lifecycle_events][:startup].clear
+    # Loaded before cron_schedule_file is set as this is the case for users of the gem
+    load 'sidekiq/cron/schedule_loader.rb'
   end
 
   describe 'Schedule file does not exist' do
     before do
       Sidekiq::Cron.configuration.cron_schedule_file = 'test/unit/fixtures/schedule_does_not_exist.yml'
-      load 'sidekiq/cron/schedule_loader.rb'
     end
 
-    it 'does not add a sidekiq lifecycle startup event' do
-      assert_nil Sidekiq::Options[:lifecycle_events][:startup].first
+    it 'does not call any sidekiq cron load methods' do
+      Sidekiq::Cron::Job.expects(:load_from_hash!).never
+      Sidekiq::Cron::Job.expects(:load_from_array!).never
+      Sidekiq::Options[:lifecycle_events][:startup].first.call
     end
 
     sidekiq_version_has_embedded = Gem::Version.new(Sidekiq::VERSION) >= Gem::Version.new('7.0.0')
@@ -27,7 +30,6 @@ describe 'ScheduleLoader' do
   describe 'Schedule is defined in hash' do
     before do
       Sidekiq::Cron.configuration.cron_schedule_file = 'test/unit/fixtures/schedule_hash.yml'
-      load 'sidekiq/cron/schedule_loader.rb'
     end
 
     it 'calls Sidekiq::Cron::Job.load_from_hash!' do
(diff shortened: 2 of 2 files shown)


tttffff:
### Previous functionality
- When `config.cron_schedule_file` has not been set: use the default `config/schedule.yml`.
- When `config.cron_schedule_file` has been set to something correct: ignore this and use the default `config/schedule.yml`.
- When `config.cron_schedule_file` has been set to something incorrect: ignore this and use the default `config/schedule.yml`.

### Functionality on this PR
- When `config.cron_schedule_file` has not been set: use the default `config/schedule.yml`.
- When `config.cron_schedule_file` has been set to something correct: use this setting.
- When `config.cron_schedule_file` has been set to something incorrect: effectively do nothing.

### Additional information
There is a potential issue when `config.cron_schedule_file` has been set to something incorrect, then the gem is updated. However, users of the gem can now [test their configuration](https://github.com/sidekiq-cron/sidekiq-cron?tab=readme-ov-file#testing-your-configuration), or run `ScheduleLoader.new.has_schedule_file?`, or simply check their configuration.

markets (approved):
Nice 👌

PR #552, opened 2025-03-24
Rework namespaces feature

viralpraxis:
The default for `available_namespaces` is now `[config.default_namespaces]`.

See https://github.com/sidekiq-cron/sidekiq-cron/issues/516

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 All notable changes to this project will be documented in this file.
 
-- **WARNING** The default value for `available_namespaces` has changed from `nil` to `[default_namespace]`. Please refer to the [migration guude](#) for further details.
+- **WARNING** The default value for `available_namespaces` has changed from `nil` to `[default_namespace]`. Please refer to the [migration guide](https://github.com/sidekiq-cron/sidekiq-cron/blob/master/README.md#migrating-to-23) for further details.
 
 ## 2.2.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 All notable changes to this project will be documented in this file.
 
+- **WARNING** The default value for `available_namespaces` has changed from `nil` to `[default_namespace]`. Please refer to the [migration guude](#) for further details.
+
 ## 2.2.0
 
 - Add support for Sidekiq v8 (https://github.com/sidekiq-cron/sidekiq-cron/pull/531, https://github.com/sidekiq-cron/sidekiq-cron/pull/538)
--- a/test/unit/job_test.rb
+++ b/test/unit/job_test.rb
@@ -6,6 +6,8 @@ describe "Cron Job" do
   before do
     Sidekiq::Cron.reset!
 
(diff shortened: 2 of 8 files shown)


markets:
@viralpraxis added small comment, but overall looks good to me (I still need to take a deeper look).

I'd like to ask for feedback and testing to all users involved in the original discussion: @Pavel-Guseynov @taylorrf @ eileen-jiang @zedtux @Yurokle.

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 All notable changes to this project will be documented in this file.
 
+- **WARNING** The default value for `available_namespaces` has changed from `nil` to `[default_namespace]`. Please refer to the [migration guude](#) for further details.
```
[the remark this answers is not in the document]
viralpraxis: Fixed by eb1e33765b5b599ad5654be0dba0300b90339aa8

test/unit/job_test.rb
```diff
@@ -1288,7 +1290,7 @@
           assert_equal 2, Sidekiq::Cron::Job.all('*').size, 'Should have 2 jobs'
 
           Sidekiq::Cron.configuration.available_namespaces = [custom_namespace]
-          assert_equal 1, Sidekiq::Cron::Job.all('*').size, 'Should have 1 job'
+          assert_equal 3, Sidekiq::Cron::Job.all('*').size, 'Should have 3 job'
```
markets, line 1293: @viralpraxis Could you please explain why this assert changes here?

zedtux:
A task to list all the namespaces from Redis records would help people being in the worst case (like me) would be really appreciated.

Otherwise this PR looks good to me.

test/unit/job_test.rb
viralpraxis: In https://github.com/sidekiq-cron/sidekiq-cron/pull/551 we made sure that the default namespace is always included in the namespaces list. Thus, `Sidekiq::Cron::Job.all('*')` **should** also include default jobs -- but that patch still relied on `Sidekiq::Cron.configuration.available_namespaces` instead of `Sidekiq::Cron::Namespace.all`. I can open another PR (as a follow-up to https://github.com/sidekiq-cron/sidekiq-cron/pull/551) if there's any confusion

viralpraxis:
> A task to list all the namespaces from Redis records would help people being in the worst case (like me) would be really appreciated.
> 
> Otherwise this PR looks good to me.

Could you clarify what you mean? I assumed you would prefer to simply set `available_namespaces` to `:auto` and keep the current (master) behavior. What do you mean by "the worst case"?

test/unit/job_test.rb
viralpraxis: In other words, [this diff](https://github.com/sidekiq-cron/sidekiq-cron/pull/552/files#diff-45ac47a9faa0840538f2cae00fdd31667a33fb6b4f94a7428d9c80bceaf8a87eL696-R696) should have been included in #551

markets: Thanks for explaining!

markets (approved):

PR #553, opened 2025-03-24
Fix invalid README.md ruby code snippet

viralpraxis:
I'd suggest adding `rubocop-md` to ensure README snippets are valid

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ gem "sidekiq-cron"
   'active_job' => true,  # enqueue job through Active Job interface
   'queue_name_prefix' => 'prefix', # Active Job queue with prefix
   'queue_name_delimiter' => '.', # Active Job queue with custom delimiter (default: '_')
-  'description' => 'A sentence describing what work this job performs'
+  'description' => 'A sentence describing what work this job performs',
   'status' => 'disabled' # default: enabled
 }
 ```


markets (approved):

viralpraxis:
What do you think about adding rubocop to the project? Seems like it might make the codebase more healthy @markets

markets:
The problem with Rubocop is to define the rules set... Some of the default ones are quite "personal taste" IMO, so in order to avoid bikeshedding we've never implemented a linter in this repo.

viralpraxis:
Yeah, that's the biggest problem with these linters -- but what about https://github.com/standardrb/standard?

markets:
I know that project exists, but I've never tried it, so I don't know if I like the presets 😅. IMO the current code base, in terms of "code style", is quite good and consistent. At this point, I'd prefer to put my efforts on other stuff instead of fighting against a linter.

viralpraxis:
Actually, perhaps you misunderstand my intention -- I did not mean to enforce specific code style (TBH I don't really can about it all and I fully agree with you that in this regard the codebase if fine), I referred to a subset of cops which have nothing to do with code style:

```console
be rubocop --plugin rubocop-performance --plugin rubocop-minitest --only Lint,Performance,Minitest
```

This currently reports 54 offenses; I have reviewed all of them and can confirm that they are all worth addressing

markets:
Then, I'd propose to send a patch fixing those offenses. I'm still not convinced to add a linter in the current workflow.

PR #555, opened 2025-03-25
Fix test-related RuboCop offenses

viralpraxis:
`bundle exec rubocop --plugin rubocop-minitest --plugin rubocop-performance --only Lint,Performance,Minitest test`

### DIFF
--- a/test/integration/performance_test.rb
+++ b/test/integration/performance_test.rb
@@ -1,10 +1,10 @@
 require './test/test_helper'
 require 'benchmark'
 
-describe 'Performance Poller' do
-  JOBS_NUMBER = 10_000
-  MAX_SECONDS = 30
+JOBS_NUMBER = 10_000
+MAX_SECONDS = 30
 
+describe 'Performance Poller' do
   before do
     Sidekiq.redis(&:flushdb)
 
--- a/test/unit/web_extension_test.rb
+++ b/test/unit/web_extension_test.rb
@@ -1,9 +1,9 @@
 require './test/test_helper'
 
+TOKEN = SecureRandom.base64(32).freeze
+
 describe 'Cron web' do
   include Rack::Test::Methods
-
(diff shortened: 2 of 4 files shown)


test/integration/performance_test.rb
```diff
@@ -2,8 +2,8 @@
 require 'benchmark'
 
 describe 'Performance Poller' do
-  JOBS_NUMBER = 10_000
-  MAX_SECONDS = 30
+  jobs_number = 10_000
+  max_seconds = 30
```
markets, line 6: Why 🤔? Conceptually, these vars are constants.

test/unit/web_extension_test.rb
```diff
@@ -3,15 +3,15 @@
 describe 'Cron web' do
   include Rack::Test::Methods
 
-  TOKEN = SecureRandom.base64(32).freeze
+  token = SecureRandom.base64(32).freeze
```
markets, line 6: Why? I don't like it, I prefer to keep the constant

test/unit/web_extension_test.rb
```diff
@@ -3,15 +3,15 @@
...
 
   def app
     Sidekiq::Web
   end
 
   before do
-    env 'rack.session', { csrf: TOKEN }
-    env 'HTTP_X_CSRF_TOKEN', TOKEN
+    env 'rack.session', { csrf: token }
+    env 'HTTP_X_CSRF_token', token
```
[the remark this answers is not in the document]
viralpraxis: Sorry, my bad. `HTTP_X_CSRF_TOKEN` should stay as is

test/unit/web_extension_test.rb
viralpraxis: Constants defined in blocks "escape" and become `Object::TOKEN`. Alternatively, we can define these constants under explicit namespace

markets (approved):

PR #557, opened 2025-04-08
Allow JSON as hash to be correctly parsed by the `Sidekiq::Cron::Job#parse_args`

francktrouillez:
Relates to:
- https://github.com/sidekiq-cron/sidekiq-cron/issues/556

This PR allows to harmonize the behavior for `Sidekiq::Cron::Job#parse_args` when using a hash directly and when a JSON representing a hash.

This allows to bring consistency with the behavior for arrays (and JSON arrays).

**This can be a considered as a breaking change if we consider people are relying on the current behavior for a JSON representing a hash for `args`.**

Indeed, `args = { "abc" => "123" }.to_json` currently gives to the worker `[ "abc", "123" ]`.

Feel free to take the relevant action for this.

### DIFF
--- a/lib/sidekiq/cron/job.rb
+++ b/lib/sidekiq/cron/job.rb
@@ -633,8 +633,7 @@ module Sidekiq
           [convert_to_global_id_hash(args)]
         when String
           begin
-            parsed_args = Sidekiq.load_json(args)
-            symbolize_args? ? symbolize_args(parsed_args) : parsed_args
+            parse_args(Sidekiq.load_json(args))
           rescue JSON::ParserError
             [*args]
           end
--- a/test/unit/job_test.rb
+++ b/test/unit/job_test.rb
@@ -1102,7 +1102,7 @@ describe "Cron Job" do
   end
 
   describe "initialize args" do
-    it "from JSON" do
+    it "from JSON as array" do
       args = {
         name: "Test",
         cron: "* * * * *",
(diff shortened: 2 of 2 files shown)


markets:
Hello @francktrouillez I think I get your point and I kind of agree, but this part is implemented that way probably from the beginning and in the method documentation, I can read:

```ruby
# Try parsing inbound args into an array.
# Args from Redis will be encoded JSON,
# try to load JSON, then failover to string array.
def parse_args(args)
...
```

So not sure if this is going to be really problematic.

francktrouillez:
@markets

I thought again about this situation, and I believe that this confusion initially comes from the fact that `args` is always assumed to be an array internally, even thought we allow non-array inputs (I know I always told "it could be another discussion", but on a second thought, it might be the root cause).

I believe that by trying to allow more and be "smart" about parsing the `args`, we actually introduced edge cases like the one I just described.

One way to prevent that would be to always expect `args` to be an array, and if the user passes a hash, we could raise an error. This way, we would avoid the confusion of having different types of arguments when calling the worker directly or through `sidekiq-cron`.
(if we don't wanna introduce breaking changes now, we could just log a warning when passing a hash, and keep the current behavior for now).

The example I gave previously would then look like this:
```ruby
class FooService
  def schedule
    Sidekiq::Cron::Job.create(
      name: "foo",
      cron: "0 0 * * *",
      class: FooWorker,
      args: [job_args]
    )
  end

  def run
    FooWorker.perform_async(job_args)
  end

  private

    def job_args
      { "my_key" => "my_value" }.to_json
    end
end

class MyWorker
  include Sidekiq::Worker

  def perform(message)
    puts "message: #{message.inspect}"
  end
end
```
where the worker could be 100% decoupled to where the arguments come from, so that the `message` is always the same.

What do you think?

markets:
Hey @francktrouillez 👋🏼 

To be 100% honest, I'm now focused on other projects and I lost track on this one. I don't fully understand all this context at this point and I have very low interest in invest my time in this PR. Maybe you're right, but I don't want to work on this at this moment. If others agree and can make proper code-review and double-check that this change doesn't break existing apps and compatibility matrix, I'd be ok to merge it.

francktrouillez:
@markets 

Thanks for the quick answer! I totally understand, I didn't notice this was closed, so I also lost track of it for a while.

> If others agree and can make proper code-review and double-check that this change doesn't break existing apps and compatibility matrix, I'd be ok to merge it.

Do you know who could potentially be involved for that?

PR #574, opened 2026-04-09
Allow to conditionally disable Sidekiq-Cron

akodkod:
Hi,

I needed to disable scheduling in some environments and found that this feature hasn't been implemented yet. I also saw that others need this too: https://github.com/sidekiq-cron/sidekiq-cron/issues/258

So I decided to implement it and contribute it for everyone.

Please let me know what you think and whether any changes are needed.

Have a good day,
Andrew

### DIFF
--- a/lib/sidekiq/cron.rb
+++ b/lib/sidekiq/cron.rb
@@ -14,6 +14,10 @@ module Sidekiq
     end
 
     class Configuration
+      # Whether Sidekiq-Cron is enabled. When set to false, the schedule file will not be loaded
+      # on startup. Defaults to true.
+      attr_accessor :enabled
+
       # The interval, in seconds, at which to poll for scheduled cron jobs.
       # This determines how frequently the scheduler checks for jobs to enqueue.
       attr_accessor :cron_poll_interval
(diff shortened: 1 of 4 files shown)


kevinrobell-st:
This would be useful. The workaround we currently use is setting the `cron_poll_interval` to zero to disable polling from some Sidekiq processes.

See https://github.com/sidekiq-cron/sidekiq-cron?tab=readme-ov-file#under-the-hood

lib/sidekiq/cron.rb
```diff
@@ -62,7 +62,12 @@ class Configuration
       # jobs that missed their schedules during the deployment. E.g., jobs that run once a day.
       attr_accessor :reschedule_grace_period
 
+      # Whether Sidekiq-Cron is enabled. When set to false, the schedule file will not be loaded
+      # on startup. Defaults to true.
+      attr_accessor :enabled
```
markets, line 67: @akodkod this is just a minor detail, but would be nice to list this at the beginning

markets:
@akodkod It would be also nice to update the section mentioned by @kevinrobell-st adding something like: "You can also disable the entire engine by ...", probably after "When running with many Sidekiq processes, the polling can add significant load to Redis. You can disable polling ...". So it will be clear that you can stop polling or disable the engine.

akodkod:
Hi @markets @kevinrobell-st

Thank you for your feedback. I've made the corresponding changes 😉

markets (approved):

PR #578, opened 2026-05-26
Allow setting `config.cron_schedule_file` to `nil` to prevent automatic loading jobs from config file

brodock:
## Context

I have a system where I've split the schedule into multiple yaml files that can be conditionally loaded. I've followed the example from: https://github.com/sidekiq-cron/sidekiq-cron#loading-jobs-from-schedule-file

Because one of the files are still called `config/schedule.yml` (this is the base one that we always want to load), it is always executed, even after I've manually loaded them before.

What we ended up doing is we set a custom name in the configuration for a file that doesn't exist. I believe this is suboptimal.

### Proposal

If `config.cron_schedule_file` is set to `nil` we never try to load it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ All configuration options:
 Sidekiq::Cron.configure do |config|
   config.enabled = false # Default is true
   config.cron_poll_interval = 10 # Default is 30
-  config.cron_schedule_file = 'config/my_schedule.yml' # Default is 'config/schedule.yml'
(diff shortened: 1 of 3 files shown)


markets:
I think that makes sense. Thanks!

markets (dismissed):

markets:
@brodock Please could you add some docs before merging?

brodock:
@markets Please let me know if the updates on README are enough

markets (approved):

